PR #3, opened 2017-10-15
Rewrite SAPI2 compatibility doc

charleswhchan:
I thought the description was a bit unclear, so I tried to rewrite it based on my understanding.

Hopefully, I conveyed the right ideas and did not misrepresent. 😃


adam-douglass:
Looks like an improvement to me.

Issue #8, opened 2017-11-20
typo, connection should be con


orenshk:
https://github.com/dwavesystems/dwave_micro_client/blob/fd86de8a2a5968dab6ce0a70b771526335255494/dwave_micro_client.py#L24

ccoffrin:
On a related note, some examples on this page use `con` and others `conn`.

http://dwave-micro-client.readthedocs.io/en/latest/

randomir:
I believe this is now fixed with #23.

PR #13, opened 2018-02-06
Client usage example/docs fix

randomir:


randomir:
This actually closes #8.

PR #30, opened 2018-02-27
Project refactor

bellert:


randomir:
Merging, but we have to fix #32.

PR #33, opened 2018-02-27
Fix open builtin mock.patch in Python < 3.5

randomir:
Note (https://docs.python.org/3/library/unittest.mock.html#patch):
> Changed in version 3.5: If you are patching builtins in a module then
you don’t need to pass `create=True`, it will be added by default.


randomir:
This fixes #32.

PR #88, opened 2018-03-31
Update example in README.rst

charleswhchan:
The original example doesn't call `client.close()` to release resources properly.
Since `client` supports context manager, using `with` will ensure resources are released when we are done with client object.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,17 +35,17 @@ Example
     from dwave.cloud import Client
 
     # Connect using the default or environment connection information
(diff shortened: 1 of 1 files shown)


randomir:
Although threads spawned by `Client` constructor will be reaped on program exit (equivalent to `client.close()`), emphasizing the context manager pattern makes sense for longer running programs.

charleswhchan:
Thanks @randomir!

PR #131, opened 2018-04-19
Restructure documentation (overall)

JoelPasvolsky:
@randomir this branch isn't ready to merge, the PR is just so you know what I'm doing.
You can see what it looks like here: http://cloudclient.readthedocs.io/en/restructure/reference/index.html
but might not be my latest and RtD is not building cli properly. My local build looks like this (I'll look into this later):
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,7 @@ instance/
 
 # Sphinx documentation
 docs/_build/
+docs/reference/generated/
 
 # PyBuilder
 target/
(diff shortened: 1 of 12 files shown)


JoelPasvolsky:
@randomir this is a general restructure. I've move the configuration examples temporarily to from_config() because that's what they demonstrate but I'll improve that and add examples to the introduction as I work my way through all the code modules.

JoelPasvolsky:
Yes, if you look at the other tools, I like to create specific examples for every function. And those can relate back to the basic to provide understanding if needed. 
For deprecated functionality I prefer to give minimum that enables existing usage but clearly discourages new usage.
I'll have a more detailed view once I get to each function

PR #139, opened 2018-04-24
Update config

JoelPasvolsky:
@randomir these are updates to config functions, I'll later add surrounding (higher level) content. Please merge when you can, if there are no big problems, it's easier for me to work on the next branch when the base is up-to-date, thanks!

### DIFF
--- a/dwave/cloud/config.py
+++ b/dwave/cloud/config.py
@@ -33,9 +33,11 @@ def get_configfile_paths(system=True, user=True, local=True, only_existing=True)
     """Return a list of local configuration files.
 
     Finds existing configuration files on the local system. Search paths
-    depend on operating system; for example, for Linux systems these might
-    include ``dwave.conf`` in the current working directory (CWD),
+    are set by homebase_ and depend on operating system; for example, for Linux systems
+    these might include ``dwave.conf`` in the current working directory (CWD),
     user-local ``.config/dwave/``, and system-wide ``/etc/dwave/``.
+
+    .. _homebase: https://github.com/dwavesystems/homebase
 
     Args:
         system (boolean, default=True):
@@ -52,7 +54,7 @@ def get_configfile_paths(system=True, user=True, local=True, only_existing=True)
 
     Returns:
         list[str]:
-            A list of configuration file paths.
+            List of configuration file paths.
 
     Examples:
         This example displays the paths of all configuration files on a Windows system
@@ -103,7 +105,7 @@ def get_configfile_path():
 
     Returns:
         str:
-            A configuration file path.
+            Configuration file path.
 
     Examples:
         This example displays the highest-priority configuration file on a Windows system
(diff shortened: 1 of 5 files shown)


dwave/cloud/config.py
```diff
@@ -255,94 +408,92 @@ def get_default_config():
...
-    If profile is not explicitly specified, selection of profile is described
-    under ``profile`` argument below.
+    """Load D-Wave Cloud Client configuration based on a configuration file.
+
+    Configuration files comply with standard Windows INI-like format,
+    parsable with Python's :mod:`configparser`. An optional ``defaults`` section
+    provides default key/values for all other sections.
+
+    Configuration values can be specified in multiple ways, ranked in the following
+    order (with 1 the highest ranked):
```
randomir: The correct order is:
- kwargs
- env
- config profile
- config defaults

But I'm not sure we need to separate config defaults from profile - it's just a profile loaded from a (set of) config file(s). And config loading mechanism uses [defaults]-values for missing profile values.

dwave/cloud/config.py
```diff
@@ -193,6 +309,42 @@ def load_profile_from_files(filenames=None, profile=None):
...
+
+        The following example code loads profile values from parsing both these files.
+
+        >>> import dwave.cloud as dc
+        >>> dc.config.load_profile_from_files(["./dwave_a.conf", "./dwave_b.conf"])   # doctest: +SKIP
+        {'client': u'sw',
+         'endpoint': u'https://url.of.some.dwavesystem.com/sapi',
+         'solver': u'EXAMPLE_2000Q_SYSTEM_A',
+         'token': u'DEF-987654321987654321987654321'}
+
```
randomir: To clarify the behaviour of profile loading function, I would add:

    >>> dc.config.load_profile_from_files(["./dwave_a.conf", "./dwave_b.conf"], profile='dw2000B')
    ...

Also, in the current example, it would be useful to note that although `dwave_b.conf` overrides `dwave_a.conf`, the function by default returns *the first* profile (here `dw2000A`), if not specified otherwise.

dwave/cloud/config.py
```diff
@@ -255,94 +408,92 @@ def get_default_config():
...
+            Path to configuration file.
 
-            If undefined (set to ``None``), the name of the config file is
-            taken from ``DWAVE_CONFIG_FILE`` environment variable.
-            If that env var is undefined or empty, the location of configuration
-            files is auto-detected, as described in :func:`load_config_from_files`.
+            If ``None``, the value is taken from ``DWAVE_CONFIG_FILE`` environment
+            variable if defined. If the environment variable is undefined or empty,
+            auto-detection searches for existing configuration files in the standard
+            directories set by homebase_.
```
randomir: It would be more technically accurate (here and elsewhere) to refer to `get_configfile_paths` instead of `homebase`, because:
- the priority (loading order) and detection of config files (system/user) is not defined in `homebase`, but in `get_configfile_paths`
- we added local file (`./dwave.conf`) to system/user dirs returned by `homebase`.

dwave/cloud/config.py
```diff
@@ -389,6 +529,46 @@ def load_config(config_file=None, profile=None, client=None,
...
+        'solver': 'EXAMPLE_2000Q_SYSTEM',
+        'token': u'ABC-123456789123456789123456789'}
+       >>> dc.config.load_config("./dwave_c.conf", profile='dw2000B', solver='Solver3')   # doctest: +SKIP
+       {'client': u'qpu',
+        'endpoint': u'https://url.of.some.dwavesystem.com/sapi',
+        'proxy': None,
+        'solver': 'Solver3',
+        'token': u'DEF-987654321987654321987654321'}
+
     """
```
randomir: It would be nice to include auto-detection example here.

PR #146, opened 2018-04-27
Update resources (first pass)

JoelPasvolsky:
@randomir I'd prefer to merge early so we don't become too out of sync. I'll afterwards go again over my new content and refine it.

### DIFF
--- a/dwave/cloud/computation.py
+++ b/dwave/cloud/computation.py
@@ -153,6 +153,27 @@ class Future(object):
 
         Returns:
             2-tuple of futures done and not_done, similar to `concurrent.futures.wait()`
+
+        Examples:
+
+        By default all sampling requests are processed asynchronously. Reading results from
+        any future object is a blocking operation.
+
+        .. code-block:: python
+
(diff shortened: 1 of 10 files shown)


dwave/cloud/computation.py
```diff
@@ -153,6 +153,27 @@ def wait_multiple(futures, min_done=None, timeout=None):
...
+
+            # We can check if a set of samples are ready without blocking
+            print(futures[0].done())
+
+            # We can wait on a single future
+            futures[0].wait()
+
+            # Or we can wait on several futures
+            dwave.cloud.computation.Future.wait_multiple(futures, min_done=3)
+
```
randomir: Heads up: mention accessing `future.samples` or `future['samples']` is a blocking operation for the first time (calls `.result()`). Next access to any of the properties is non-blocking.

Also, `as_completed` is modeled after `concurrent.futures.as_completed` (has same semantics as much as possible). `as_completed` is usually preferred over `wait_multiple` since it is a blocking iterable of resolved futures.

Issue #147, opened 2018-04-30
Make Solver.retrieve_problem() a private function


JoelPasvolsky:
Radomir Stevanovic [12:13 PM]
As far as understand it so far - user should never call that function directly.
currently it's used only for testing
I can make it a private function

PR #149, opened 2018-04-30
Update solver class & methods

JoelPasvolsky:
@randomir without retrieve_problem() -- thanks for closing issue #147 so quickly!

### DIFF
--- a/dwave/cloud/solver.py
+++ b/dwave/cloud/solver.py
@@ -156,9 +156,9 @@ class Solver(object):
             ...     solver = client.get_solver()
             ...     u, v = next(iter(solver.edges))
             ...     computation = solver.sample_ising({u: -1, v: 1},{}, num_reads=5)   # doctest: +SKIP
+            ...     for i in range(5):
+            ...         print(computation.samples[i][u], computation.samples[i][v])
             ...
-            >>> for i in range(5):    # doctest: +SKIP
-            ...     print(computation.samples[i][u], computation.samples[i][v])
             ...
             (1, -1)
             (1, -1)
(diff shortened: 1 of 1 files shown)


dwave/cloud/solver.py
```diff
@@ -127,22 +144,66 @@ def sample_ising(self, linear, quadratic, **params):
...
+            simple :term:`Ising` problem (opposite linear biases on two coupled qubits), and samples
+            5 times.
+
+            >>> from dwave.cloud import Client
+            >>> with Client.from_config() as client:
+            ...     solver = client.get_solver()
+            ...     u, v = next(iter(solver.edges))
+            ...     computation = solver.sample_ising({u: -1, v: 1},{}, num_reads=5)   # doctest: +SKIP
+            ...
+            >>> for i in range(5):    # doctest: +SKIP
```
randomir: It's important to do all communication with SAPI under a live `Client` context.
On context closing (exit), workers responsible for checking problem status on SAPI and downloading of results are terminated -- and pending futures are *never* resolved.

The `for` loop in this example must be under `with` block. It wouldn't have to be only if future was resolved within the context, with e.g. `samples = computation.samples`, or `result = computation.result()`.

dwave/cloud/solver.py
```diff
@@ -127,22 +144,66 @@ def sample_ising(self, linear, quadratic, **params):
...
+            samples 5 times.
+
+            >>> from dwave.cloud import Client
+            >>> with Client.from_config() as client:  # doctest: +SKIP
+            ...     solver = client.get_solver()
+            ...     u, v = next(iter(solver.edges))
+            ...     Q = {(u, u): -1, (u, v): 0, (v, u): 2, (v, v): -1}
+            ...     computation = solver.sample_qubo(Q, num_reads=5)
+            ...
+            >>> for i in range(5):     # doctest: +SKIP
```
randomir: Same as above.

PR #152, opened 2018-05-01
Update computation (without occurrences property)

JoelPasvolsky:
@randomir: I'll talk to you tomorrow about issue#151

### DIFF
--- a/dwave/cloud/computation.py
+++ b/dwave/cloud/computation.py
@@ -168,35 +168,47 @@ class Future(object):
     def wait_multiple(futures, min_done=None, timeout=None):
         """Wait for multiple :class:`Future` objects to complete.
 
-        A blocking call that uses an event object to emulate multi-wait for Python.
+        Blocking call that uses an event object to emulate multi-wait for Python.
 
         Args:
-            futures (list of Future): list of objects to wait on
-            min_done (int): Stop waiting when this many results are ready
-            timeout (float): Maximum number of seconds to wait, `None` for indefinite wait.
+            futures (list of Future): List of :class:`Future` objects to await.
+            min_done (int): Minimum required completions to end the waiting. The wait is terminated
+                when this number of results are ready. If None, waits for all the
+                :class:`Future` objects to complete.
+            timeout (float): Maximum number of seconds to await completion. If None,
+                awaits indefinitely.
 
         Returns:
-            2-tuple of futures done and not_done, similar to `concurrent.futures.wait()`
+            Two-tuple of :class:`Future` objects completed and not completed,
+                similar to `concurrent.futures.wait()` returned two-tuple of 'done' and
+                'not_done' sets.
 
         Examples:
+            This example creates a solver using the local system’s default D-Wave Cloud Client
+            configuration file, submits a simple QUBO problem to a remote D-Wave resource 3 times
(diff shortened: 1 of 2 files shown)


JoelPasvolsky:
@randomir: doesn't look to me that Travis error is due to content added in this PR

dwave/cloud/computation.py
```diff
@@ -1,5 +1,15 @@
...
+
+Your solver instantiates a :class:`Future` object for its calls, via D-Wave Sampler API
+(SAPI) servers, to the remote resource.
+
+You can interact with the :class:`Future` object for pending calls: monitoring problem status,
+waiting for and retrieving results, cancelling enqueued jobs, etc.
+
+Some :class:`Future` methods are blocking.
+
 """
```
randomir: I would also clarify the meaning behind "computation" name. Through computation module we prove to users an interface to the running (or completed) computation (sampling on QPU or sw) that's happening on the remote side. `Future` object allows them to query the current state of that computation and read the result when available.

dwave/cloud/computation.py
```diff
@@ -25,21 +35,37 @@
 
 @functools.total_ordering
 class Future(object):
```
randomir: I think it's important to note that users should *never* create `Future` objects manually. They would be "dead", since problem submittal is done in Client, initiated by Solver. Messy for now.

dwave/cloud/computation.py
```diff
@@ -220,13 +258,49 @@ def wait_multiple(futures, min_done=None, timeout=None):
...
+            ...                solver.sample_qubo(Q, num_reads=50),
+            ...                solver.sample_qubo(Q, num_reads=10)]   # doctest: +SKIP
+            >>> for tasks in dc.computation.Future.as_completed(computation, timeout=10)
+            ...     print(tasks.timing)   # doctest: +SKIP
+            ...
+            {u'total_real_time': 17318, ... u'qpu_readout_time_per_sample': 123}
+            {u'total_real_time': 10816, ... u'qpu_readout_time_per_sample': 123}
+            {u'total_real_time': 26285, ... u'qpu_readout_time_per_sample': 123}
+            >>> # Snipped above response for brevity
+            >>> client.close()
```
randomir: We can also add example for the task from `wait_multiple` above (get the result of the first computation finished):
```
from dwave.cloud.computation import Future
...
first = next(Future.as_completed(computations))
print(first.timing)
```

dwave/cloud/computation.py
```diff
@@ -237,22 +311,91 @@ def as_completed(fs, timeout=None):
...
+
+        Blocking call that waits for a :class:`Future` object to complete.
 
         Args:
-            timeout (float): Maximum number of seconds to wait
+            timeout (float, optional, default=None): Maximum number of seconds to await completion.
+                If None, waits indefinitely.
+
+        Returns:
+            Boolean: True if solver recieved a response.
```
randomir: typo

JoelPasvolsky: Fixed

dwave/cloud/computation.py
JoelPasvolsky: Done

dwave/cloud/computation.py
JoelPasvolsky: Added

PR #155, opened 2018-05-03
Update CLI

JoelPasvolsky:


docs/reference/intro.rst
```diff
@@ -88,7 +88,19 @@ One config file can contain multiple profiles, each defining a separate
 Interactive CLI Configuration
 -----------------------------
 
-TODO
+As part of the installation of the D-Wave Cloud Client package, a `dwave` executable
+is installed; for example, in a virtual environment it might be installed as
+`<virtual_environment>\\Scripts\\dwave.exe`. Running this file from your system's
```
randomir: Note: *only* `dwave config create` command is actually interactive.

JoelPasvolsky: Yes, I feel comfortable with the distinction between "an interactive command line interface (CLI) that guides you through setting up a D-Wave Cloud Client configuration file" on one hand, and "also provides additional helpful functionality" on the other.

PR #168, opened 2018-05-10
Bugfix problem encoding issue 165

randomir:
Fix #151, #165, #167.

Issue #151, opened 2018-05-01
computation.occurrences() output not as expected


JoelPasvolsky:
I set QPU parameter answer_mode=histogram and submit a problem with several ground states, I expected to see the result returned in a histogram with ~4 occurrences buckets, similar to what results through SAPI:
![image](https://user-images.githubusercontent.com/34041130/39498504-5df90e42-4d5d-11e8-85fb-bad595f0ae64.png)

The resultant samples have repetitions (e.g. 3rd and 4th) but occurrences are always 1:

```
In [690]: with Client.from_config() as client:
     ...:   solver = client.get_solver()
     ...:   solver.parameters['answer_mode']='histogram'
     ...:   quad = {(16, 20): -1, (17, 20): 1, (16, 21): 1, (17, 21): 1}
     ...:   computation = solver.sample_ising({}, quad, num_reads=20)
     ...:   for i in range(len(computation.occurrences)):
     ...:       print(computation.samples[i][16],computation.samples[i][17], computation.samples[i][20], computation.sa
     ...: mples[i][21], ' --> ', computation.energies[i], computation.occurrences[i])
     ...:
(-1, 1, -1, 1, ' --> ', -2.0, 1)
(-1, -1, -1, 1, ' --> ', -2.0, 1)
(1, -1, 1, -1, ' --> ', -2.0, 1)
(1, -1, 1, -1, ' --> ', -2.0, 1)
(1, -1, 1, -1, ' --> ', -2.0, 1)
(1, -1, 1, 1, ' --> ', -2.0, 1)
(-1, 1, -1, -1, ' --> ', -2.0, 1)
(1, -1, 1, 1, ' --> ', -2.0, 1)
(1, -1, 1, 1, ' --> ', -2.0, 1)
(-1, -1, 1, 1, ' --> ', -2.0, 1)
(1, -1, 1, -1, ' --> ', -2.0, 1)
(-1, 1, -1, -1, ' --> ', -2.0, 1)
(1, 1, -1, -1, ' --> ', -2.0, 1)
(-1, -1, -1, 1, ' --> ', -2.0, 1)
(-1, 1, -1, 1, ' --> ', -2.0, 1)
(-1, -1, 1, 1, ' --> ', -2.0, 1)
(1, 1, -1, -1, ' --> ', -2.0, 1)
(1, 1, 1, -1, ' --> ', -2.0, 1)
(1, 1, 1, -1, ' --> ', -2.0, 1)
[...]

PR #179, opened 2018-05-14
Update computation (with occurrences property)

JoelPasvolsky:
@randomir: updaed with occurrences property for your fix of issue #151 via PR #168.

### DIFF
--- a/dwave/cloud/computation.py
+++ b/dwave/cloud/computation.py
@@ -462,7 +462,7 @@ class Future(object):
         to this property is non-blocking.
 
         Returns:
-            list or numpy matrix of doubles: Energies for each set of samples.
+            list or NumPy matrix of doubles: Energies for each set of samples.
 
         Examples:
             This example creates a solver using the local system's default D-Wave Cloud Client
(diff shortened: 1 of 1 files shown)


dwave/cloud/computation.py
```diff
@@ -526,7 +526,35 @@ def occurrences(self):
...
+            This example creates a solver using the local system's default D-Wave Cloud Client
+            configuration file, submits a simple Ising problem with several ground states to a
+            remote D-Wave resource for 20 samples, and prints the returned results, which
+            are ordered as a histogram. The problem's ground states tend to recur frequently,
+            and so those solutions have `occurrences` greater than 1.
+
+            >>> from dwave.cloud import Client
+            >>> with Client.from_config() as client:  # doctest: +SKIP
+            ...     solver = client.get_solver()
+            ...     solver.parameters['answer_mode']='histogram'
```
randomir: Instead of modifying solver parameters, supplying the value as kwarg to sampling method seems cleaner:
```
computation = solver.sample_ising({}, quad, num_reads=500, answer_mode='histogram')
```
Can we use that instead?

PR #191, opened 2018-06-07
Change Client for better .solvers() experience and more

randomir:
Use nicer Solver object representation (e.g. `Solver(id='DW_2000Q_2')`
instead of `<dwave.cloud.solver.Solver object at 0x7f6c0f4b06d8>`).

Make `Client.from_config` more consistent with `Client` constructor
by defaulting to creating an instance of the actual `Client` class used
(generic in `client`, QPU in `qpu`, or software in `sw`).

### DIFF
--- a/dwave/cloud/client.py
+++ b/dwave/cloud/client.py
@@ -374,8 +374,8 @@ class Client(object):
         config.update(kwargs)
 
         from dwave.cloud import qpu, sw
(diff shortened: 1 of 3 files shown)


randomir:
@JoelPasvolsky, @arcondello: not much to review -- just a heads up, we have a new default behaviour in cloud client's `from_config`. I would say more consistent and expected.

PR #222, opened 2018-07-27
Add timeout

randomir:


dwave/cloud/client.py
```diff
@@ -383,7 +383,7 @@ def from_config(cls, config_file=None, profile=None, client=None,
         return _clients[_client](**config)
 
     def __init__(self, endpoint=None, token=None, solver=None, proxy=None,
-                 permissive_ssl=False, **kwargs):
+                 permissive_ssl=False, timeout=60, **kwargs):
```
arcondello: Why 60?

randomir: Just a reasonable default. Since users are not expecting a timeout, we don’t want to timeout too often. But it’s customizable for power users.

tests/test_mock_solver_loading.py
```diff
@@ -228,6 +230,12 @@ def handle(path, *args, **kwargs):
 endpoint = file-alpha-url
 token = file-alpha-token
 solver = alpha-solver
+
+[custom]
+endpoint = http://httpbin.org/delay/10
+token = 123
+permissive_ssl = True
+timeout = 15
```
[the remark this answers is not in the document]
randomir: No env vars for all these additional params, just basic ones. Open for discussion if we want to introduce a general naming scheme for arg/env/conf.

PR #242, opened 2018-09-17
Add SolverOfflineError exception

randomir:


dwave/cloud/exceptions.py
```diff
@@ -14,6 +14,9 @@ class SolverError(Exception):
 class SolverFailureError(SolverError):
     """An exception raised when there is a remote failure calling a solver."""
```
JoelPasvolsky: Raised when calling a remote solver fails.

dwave/cloud/exceptions.py
```diff
@@ -14,6 +14,9 @@ class SolverError(Exception):
...
 
+class SolverOfflineError(SolverError):
+    """Action attempted on an offline solver."""
```
JoelPasvolsky: Raised when a requested solver is in an offline state

dwave/cloud/exceptions.py
```diff
@@ -14,6 +14,9 @@ class SolverError(Exception):
...
+
 class SolverAuthenticationError(SolverError):
     """An exception raised when there is an authentication error."""
```
JoelPasvolsky: Raised when user authentication fails.

dwave/cloud/client.py
```diff
@@ -901,7 +901,12 @@ def _handle_problem_status(self, message, future):
                 future._set_error(CanceledFutureError())
             else:
                 # Return an error to the future object
-                future._set_error(SolverFailureError(message.get('error_message', 'An unknown error has occurred.')))
+                errmsg = message.get('error_message', 'An unknown error has occurred.')
+                if 'solver is offline' in errmsg.lower():
+                    err = SolverOfflineError(errmsg)
+                else:
+                    err = SolverFailureError(errmsg)
```
JoelPasvolsky: How come you're not checking for the other known errors, like authentication?

randomir: They are checked for before.

PR #260, opened 2018-10-09
Show 'Client.solvers' in docs

randomir:


docs/reference/resources.rst
```diff
@@ -43,6 +43,7 @@ Methods
    :toctree: generated
 
    client.Client.from_config
+   client.Client.solvers
```
JoelPasvolsky: I already see this https://docs.ocean.dwavesys.com/projects/cloud-client/en/latest/reference/resources.html
image

dwave/cloud/client.py
```diff
@@ -737,9 +737,9 @@ def get_solver(self, name=None, features=None, refresh=False):
 
             features (dict, optional):
                 Dictionary of features this solver has to have. For a list of
-                feature names and values, see: :meth:`Client.solvers`. Specifying solver
-                name overrides features. To include (possibly partial) name match requirement,
-                use the ``name`` feature inside ``features`` dictionary.
+                feature names and values, see: :meth:`~dwave.cloud.client.Client.solvers`.
+                Specifying solver name overrides features. To include (possibly partial)
```
JoelPasvolsky: Specifying the 'name' parameter overrides the 'feature' parameter.

dwave/cloud/client.py
```diff
@@ -737,9 +737,9 @@ def get_solver(self, name=None, features=None, refresh=False):
...
+                name match requirement, use the ``name`` feature inside ``features`` dictionary.
```
JoelPasvolsky: To require a (full or partial) name match, use the 'features` parameter to specify a value for its 'name' key.

docs/reference/resources.rst
randomir: Yes, because I already merged this PR. :)

dwave/cloud/client.py
randomir: Thanks, will change in the next PR.

dwave/cloud/client.py
randomir: Changed! Tnx.

PR #261, opened 2018-10-10
Add 'anneal_schedule' Solver feature

randomir:


JoelPasvolsky:
@randomir, I think for the anneal_schedule feature, we need at least 4 points:

```
In [11]: solver.has_anneal_schedule
Out[11]: True
In [12]: print("max anneal schedule points: {}".format(solver.properties["max_anneal_schedule_points"]))
max anneal schedule points: 2
```

randomir:
@JoelPasvolsky, no we don't. Sorry, but these are separate features.

PR #263, opened 2018-10-13
Arbitrary property filtering in solver selector

randomir:


dwave/cloud/client.py
```diff
@@ -630,11 +631,15 @@ def get_solvers(self, refresh=False):
...
             return self._solvers
 
-    def solvers(self, name=None, qpu=None, software=None, vfyc=None,
-                flux_biases=None, anneal_schedule=None,
-                num_qubits=None, online=True, refresh=False):
+    def solvers(self, name=None, qpu=None, software=None, online=True,
+                refresh=False, **filters):
         """Returns a filtered list of solvers handled by this client.
 
+        Apart from a few inferred features (`name`, `qpu`, `software`, and
```
JoelPasvolsky: "inferred" ?
Maybe drop this paragraph and move the "similar to Django QuerySet..." to the description of **filters. Having some explicit parameters in addition to a more general catch-all argument should not be confusing to users.

dwave/cloud/client.py
```diff
@@ -646,24 +651,62 @@ def solvers(self, name=None, qpu=None, software=None, vfyc=None,
...
-                Should solver accept anneal schedule?
-
-            num_qubits (int/[int,int], default=None):
-                What's the range (or exact number) of qubits solver should handle?
-
             online (bool, default=True):
                 Should solver be online?
 
             refresh (bool, default=False):
                 Force refresh cached list of solvers/properties
```
JoelPasvolsky: Here I'd place the **filters parameter with its explanation

dwave/cloud/client.py
```diff
@@ -646,24 +651,62 @@ def solvers(self, name=None, qpu=None, software=None, vfyc=None,
...
 
+        Filtering forms:
+            <parameter> (bool),
+            <parameter>__available (bool),
+            <parameter>__<operator> (object <value>):
+                Ensures solver supports `parameter`. General operator form can
+                be used, but that usually doesn't make sense for parameters,
+                since values are human-readable descriptions. Default operator
+                if `available`.
```
JoelPasvolsky: Did you mean "if" and not "is" here?

dwave/cloud/client.py
```diff
@@ -646,24 +651,62 @@ def solvers(self, name=None, qpu=None, software=None, vfyc=None,
...
+            <property>__eq (bool),
+            <property>__<operator> (object <value>):
+                Ensures the value of solver's `property`, after applying
+                `operator` equals the righthand side `value`. Default operator
+                is `eq`.
+
+        Operators:
+            available, lt, lte, gt, gte, eq, within, contains, regex
+
+        Common solver parameters:
```
JoelPasvolsky: Reversing roles here for a moment: why not save the docstring lines of these "Common solver parameters" and "Common solver properties", and let the examples, which use these same parameters/properties  speak for themselves?

dwave/cloud/client.py
```diff
@@ -675,17 +718,28 @@ def solvers(self, name=None, qpu=None, software=None, vfyc=None,
             class.
 
         Examples:
-            Get a solver not based on VFYC, with 2048 qubits::
+            Get all solvers based on VFYC, with 2048 qubits::
+
+                solver = client.solvers(vfyc=True, num_qubits=2048)
+                solver = client.solvers(vfyc__eq=True, num_qubits__eq=2048)
```
JoelPasvolsky: We'll get support questions on why two code lines are needed. I suggest instead one line with mixed syntax:
solver = client.solvers(vfyc=True, num_qubits__eq=2048)

dwave/cloud/client.py
```diff
@@ -696,36 +750,87 @@ def solvers(self, name=None, qpu=None, software=None, vfyc=None,
...
                 solver = client.solvers(online=True, name='solver1|solver2')[0]
+                solver = client.solvers(online=True, chip_id__regex='solver[12]')[0]
+
+            Get all solvers that contain a coupler ``(0, 128)``::
+
+                solver = client.solvers(couplers__contains=[0, 128])
+
+            Get all solvers that support 1000 for ``num_reads``:
+
+                solver = client.solvers(num_reads_range__covers=1000)
```
JoelPasvolsky: "covers" is not in the above list: "available, lt, lte, gt, gte, eq, within, contains, regex "

dwave/cloud/client.py
```diff
@@ -696,36 +750,87 @@ def solvers(self, name=None, qpu=None, software=None, vfyc=None,
...
         """
 
-        def predicate(solver):
+        def covers_op(prop, val):
+            """Does LHS `prop` (range) fully cover RHS `val` (range or item)?"""
```
JoelPasvolsky: Verify that range `prop` fully covers range or item `val`.

dwave/cloud/client.py
```diff
@@ -696,36 +750,87 @@ def solvers(self, name=None, qpu=None, software=None, vfyc=None,
...
+            if isinstance(val, (list, tuple)) and len(val) == 2:
+                # val range within prop range?
+                rlo, rhi = min(val), max(val)
+                return llo <= rlo and lhi >= rhi
+            else:
+                # val item within prop range?
+                return llo <= val <= lhi
+
+        def coveredby_op(prop, val):
+            """Is LHS `prop` (range or item) fully covered by RHS `val` (range)?"""
```
JoelPasvolsky: Verify that range or item `prop` () is fully covered by range `val`.

dwave/cloud/client.py
randomir: Already fixed, thanks. (Seems like you haven't reviewed the latest version of code)

dwave/cloud/client.py
randomir: You're right, I need to slash all those examples a little bit :)

dwave/cloud/client.py
randomir: Just wanted to emphasize either works. I'll add a "or".

dwave/cloud/client.py
randomir: The list is already fixed (in latest commit).

dwave/cloud/client.py
randomir: Resolved now.

PR #276, opened 2018-11-03
Solvers getters refactor

randomir:


dwave/cloud/utils.py
```diff
@@ -247,3 +247,64 @@ class CLIError(Exception):
     def __init__(self, message, code):
         super(CLIError, self).__init__(message)
         self.code = code
+
+
+class cached(object):
```
arcondello: might consider a more descriptive name `memoization_cache` or something. I am not sure how I would interpret without this context

```
@cached()
def f():
    pass
```

randomir: Good point, `@cached` might be unclear to some. But memoization this is not - because the result is expected not to be deterministic - hence the cache timeout/maxage parameter.

It's internalish function anyway, so in lack of a better name, I'm keeping `cached` for now.

Issue #272, opened 2018-10-31
More descriptive error message in `dwave ping` when certificate verification fails


randomir:
HTTPS certificates used on non-official SAPI endpoints might not be issues/signed by publicly recognized CA (think proxies and customers with on site installations of a D-Wave system).

In that case, we can advise users to install their local CA to a system-wide certificate store, or if they are sure they want to forgo all transport-layer security, they can use the `permissive_ssl` flag.

PR #277, opened 2018-11-06
Check for (private) cert verify fail; direct user to KB article

randomir:


randomir:
Yes, by grep I mean string search :)

randomir:
Closes #272.

PR #283, opened 2018-12-19
Update feature-based solver selection

JoelPasvolsky:


docs/intro.rst
```diff
@@ -26,7 +26,7 @@ It's recommended you set up your D-Wave Cloud Client configuration through the
 
 As described in the :std:doc:`Using a D-Wave System <oceandocs:overview/dwavesys>` section
 of Ocean Documentation, for your code to access remote D-Wave compute resources, you must
-configure communication through SAPI; for example, your code needs the SAPI URL and your API
+configure communication through SAPI; for example, your code needs your API
```
randomir: Should we mention here that API endpoint URL can be changed, to accommodate for example on premise installs of QCs?

docs/intro.rst
```diff
@@ -89,23 +89,25 @@ through auto-detection as the default configuration, contains the following
...
 
           [defaults]
-          endpoint = https://url.of.some.dwavesystem.com/sapi
-          client = qpu
-
-          [dw2000]
-          solver = DW_2000Q_1
           token = ABC-123456789123456789123456789
 
+          [firstAvailableQPU]
```
randomir: I would avoid the lower camel case (non-standard Python) and use `first-available-qpu`.

docs/intro.rst
```diff
@@ -89,23 +89,25 @@ through auto-detection as the default configuration, contains the following
...
+          solver = {"qpu": true}
+
           [software]
           client = sw
           solver = c4-sw_sample
           token = DEF-987654321987654321987654321
           proxy = http://user:pass@myproxy.com:8080/
 
-You can instantiate clients for a D-Wave 2000Q QPU and a CPU with::
+          [backupDwave2000Q]
```
randomir: `backup-dwave-2000q`?

dwave/cloud/client.py
```diff
@@ -106,10 +106,10 @@ class Client(object):
     Examples:
         This example directly initializes a :class:`~dwave.cloud.client.Client`.
         Direct initialization uses class constructor arguments, the minimum being
-        values for `endpoint` and `token`.
+        a value for `token`.
 
         >>> from dwave.cloud import Client
-        >>> client = Client(endpoint='https://cloud.dwavesys.com/sapi', token='secret')
+        >>> client = Client(token='secret')
```
randomir: Not sure if this is the right venue for this example, but I would be useful to note that for such "minimal" client, `client.get_solver()` will return "the first" (by id) available online solver. It might happen that's a software solver.

randomir: Also, to be more specific, user will have to say `client.get_solver(<solver-def-kwargs>)`.

dwave/cloud/client.py
```diff
@@ -530,119 +530,111 @@ def _fetch_solvers(self, name=None):
...
+                Force refresh of cached list of solvers/properties.
+            filters:
+                See `Filtering forms` and `Operators` below.
+
+        Solver filters are defined, similarly to Django QuerySet filters, with
+        keyword arguments of form `<name>__<operator>=<value>`. Each `<operator>`
+        is a predicate (boolean) function that acts on two arguments: value of feature
+        `<name>` and the required `<value>`.
+
+        Feature `<name>` can be:
```
randomir: Using interpreted quote in sphinx will not resolve <name>, since it's not defined anywhere... That's why I like to use double ticks, or alternatively italics for such symbols.

dwave/cloud/config.py
```diff
@@ -103,30 +103,29 @@
...
-        client = qpu
-
-        [dw2000A]
-        solver = EXAMPLE_2000Q_SYSTEM_A
         token = ABC-123456789123456789123456789
 
+        [primary-qpu]
+        solver = {"qpu": true}
+
         [sw_solver]
```
randomir: Let's make profile names consistent. Now we have `a-b`, `a_b` and `aB`.

dwave/cloud/config.py
```diff
@@ -724,21 +722,21 @@ def load_config(config_file=None, profile=None, client=None,
             Config file parse failed.
 
     Examples
-        Additional examples are given in :mod:`dwave.cloud.config`.
+        This example loads the configuration from an auto-detected configuration file
+        in the home directory of a Windows system user.
+
+        >>> import dwave.cloud as dc
+        >>> dc.config.load_config()
```
randomir: Can we avoid the awful `dc` module name?
`from dwave.cloud.config import load_config, get_configfile_paths`, or `from dwave.cloud import config`?

docs/intro.rst
JoelPasvolsky: This is very much the entry point to this tool, and I'm thinking that for now the number of users who might benefit from that info is small compared to those for whom it will be a distraction.

docs/intro.rst
JoelPasvolsky: I'll make all these changes

dwave/cloud/client.py
JoelPasvolsky: Soon this will be covered by "By default, solvers are sorted by average load."

dwave/cloud/client.py
JoelPasvolsky: We differ there. I find that in RTD the double quotes are too prominent and resemble the clickable links to other methods/classes, and that will be frustrating for users:
image

dwave/cloud/config.py
JoelPasvolsky: Done

dwave/cloud/config.py
JoelPasvolsky: Yes, I've been slowly hunting these down. I'll do that next time I'm in this repo, I'll get the current fixes in for you now.

Issue #297, opened 2019-01-11
'dwave solvers' not ignoring solver definition from the config


randomir:
`dwave solvers` should fetch all online solvers by default, regardless of the `solver` definition in any of the configuration channels (file or env in this case).

Currently, it doesn't.

PR #298, opened 2019-01-11
Fix zero-length solver def json parsing

randomir:
Essentially, ignore `solver =` lines in config file, or 'DWAVE_SOLVER='
env var values.


randomir:
Close #297.

PR #321, opened 2019-07-30
Support bqm sampling on unstructured solvers

randomir:


dwave/cloud/coders.py
```diff
@@ -237,4 +242,59 @@ def decode_qp_numpy(msg, return_matrix=True):
...
 
+    # include problem type
+    if 'type' in msg:
+        result['problem_type'] = msg['type']
+
+    return result
+
+
+def encode_problem_as_bq(problem, compress=False):
+    """Encode the binary quadratic problem for submission the `bq` data format.
```
JoelPasvolsky: Accidentally noticed what I think is a missing "in". Also, we usually use "binary quadratic model" or problem but not "binary quadratic problem", typically the BQM represents a reformulation of he user's problem, so I think it's helpful to keep that distinction.

dwave/cloud/coders.py
```diff
@@ -237,4 +242,59 @@ def decode_qp_numpy(msg, return_matrix=True):
...
+
+    Args:
+        problem (:class:`~dimod.BinaryQuadraticModel`):
+            Binary quadratic model.
+
+    Returns:
+        encoded submission dictionary
```
JoelPasvolsky: While I'm there anyway: "Encoded submission dict."

randomir: @JoelPasvolsky, feel free to make a(nother) pass through the cloud client docs. I refactored docs in functions that I touched in this PR, but most of the documentation hasn't been updated since the times of microclient.

dwave/cloud/coders.py
randomir: Fixed the "in" typo, thanks.

Re: bqp, I see the term is used elsewhere, and it was heavily used in the qbsolv/microclient era. Even the encoding format on SAPI is called `qp` (quadratic problem).

If we want to use the new terminology in CC, we should consistently change it. And that's a documentation refactor I mentioned yesterday.

randomir:
Currently failing on Linux because `dimod==0.8.15` not available on Linux (see https://github.com/dwavesystems/dimod/issues/514).

dwave/cloud/coders.py
```diff
@@ -237,4 +242,60 @@ def decode_qp_numpy(msg, return_matrix=True):
...
+    """
+
+    import zlib
+    import json
+
+    serialized_bqm = problem.to_serializable(use_bytes=False)
+
+    if compress:
+        # note: compression scheme will change, and preferably move to transport layer
+        compressed_bqm = zlib.compress(bytes(json.dumps(serialized_bqm), "ascii"))
```
oneklc: oh my!!
you sure you wanna leave this in?
I agree with using the transport layer instead of this also

randomir: I though you want to use application layer compression, at least during dev. Are you not? I can remove it if you're sure.

randomir:
[Dimod deploy issue](https://github.com/dwavesystems/dimod/issues/514) fixed, tests are passing now. Force pushed to trigger rebuild.

dwave/cloud/computation.py
```diff
@@ -594,33 +674,79 @@ def occurrences(self):
...
+
+        result = self._load_result()
+        if 'sampleset' in result:
+            return result['sampleset']
+
+        # construct sampleset from available data
+        try:
+            import dimod
+        except ImportError:
+            raise RuntimeError("Can't construct SampleSet without dimod. "
```
arcondello: Maybe keep this as an `ImportError` rather than  `RuntimeError`?

dwave/cloud/solver.py
```diff
@@ -200,6 +223,166 @@ def is_online(self):
...
+        try:
+            import dimod
+        except ImportError: # pragma: no cover
+            raise RuntimeError("Can't use solver of type 'bqm' without dimod. "
+                               "Re-install the library with 'bqm' support.")
+
+        bqm = dimod.BinaryQuadraticModel.from_qubo(qubo)
+        return self.sample_bqm(bqm, **params)
+
+    def sample_bqm(self, bqm, **params):
```
wbernoudy: Should you also raise a RuntimeError like the other sample methods explaining this needs dimod? Maybe it's not necessary because it explains what the `bqm` param is supposed to be in the docstring...

randomir: Strictly speaking, for this method, dimod is not required. That is, dimod calls are not used anywhere on the "submit-this-bqm-to-sapi" code path. The only requirement is for the input `bqm` object to have `.to_serializeable()`.

That's unlike other sample methods here, in which we use dimod to construct a BQM object.

dwave/cloud/computation.py
randomir: Yes, that's a good point. I considered using `ImportError`, but re-throwing `ImportError` with a different message (having in mind this exception is used (almost?) exclusively by the `import` statement) seems wrong. Or at least not "obvious way to do it".

Due to exception nesting, user will be able to see the whole exception stack trace anyway, and see that the `RuntimeError` was caused by a failing import.

As a Python user I would be surprised to see `ImportError` raised because of another `ImportError` (like `imp` or `importlib` failed importing something...)

If you're concerned about handling this case programmatically, then subclassing (of `ImportError`) would probably be the best way. But I don't think we should overthink this because dimod might very well become dependency of the client in future.

PR #331, opened 2019-10-18
Multipart upload

randomir:


dwave/cloud/client.py
```diff
@@ -147,6 +155,8 @@ class Client(object):
 
     # Number of worker threads for each problem processing task
     _SUBMISSION_THREAD_COUNT = 5
+    _UPLOAD_PROBLEM_THREAD_COUNT = 1
+    _UPLOAD_PART_THREAD_COUNT = 10
```
arcondello: Out of curiosity, where do these numbers come from?

dwave/cloud/utils.py
```diff
@@ -130,6 +145,25 @@ def uniform_get(sequence, index, default=None):
         return sequence[index] if index < len(sequence) else default
 
 
+def qubo_to_ising(qubo):
```
arcondello: You might want to rename this to distinguish it from say [qubo_to_ising](https://github.com/dwavesystems/dimod/blob/master/dimod/utilities.py#L217) which does the vartype conversion as well.

dwave/cloud/utils.py
```diff
@@ -249,6 +283,32 @@ def send(self, *args, **kwargs):
         return super(TimeoutingHTTPAdapter, self).send(*args, **kwargs)
 
 
+# Note: BaseUrlSession is taken from https://github.com/requests/toolbelt. This
```
arcondello: Maybe add "under an apache license"

dwave/cloud/client.py
randomir: Just initial estimates. Actual workers are extremely IO-bound, so we might as well just resort to [the default](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) in the standard library (which varies across versions; and in Py3 actual number of workers is adaptive but less then the given max).

PR #345, opened 2020-01-28
Add `Client.get_problem` method

### LINKED ISSUE #186: Enable results download by problem ID
For old/out-of-context problems, referred by ID, enable futures to be recreated and results downloaded. Something like this, but user-friendly:
```
>>> from dwave.cloud.computation import Future
>>> from dwave.cloud import Client
>>> c = Client.from_config()
>>> s = c.get_solver()
>>> 
>>> f = Future(s, '85fa54d1-0055-46f0-9f45-e8b99a9d857e', False, {})
>>> 
>>> c._load(f)
>>> f.done()
True
>>> f.result()
{'samples': [[-1, 0, 0, ....
```

arcondello:
An attempt to allow download by problem id.

Closes #186

### DIFF
--- a/dwave/cloud/client.py
+++ b/dwave/cloud/client.py
@@ -68,6 +68,7 @@ import six
 
 from dwave.cloud.package_info import __packagename__, __version__
 from dwave.cloud.exceptions import *
+from dwave.cloud.computation import Future
 from dwave.cloud.config import load_config, legacy_load_config, parse_float
 from dwave.cloud.solver import Solver, available_solvers
 from dwave.cloud.concurrency import PriorityThreadPoolExecutor
@@ -646,6 +647,21 @@ class Client(object):
 
         return solvers
 
+    def get_problem(self, id_):
+        """Retrieve a problem by id.
+        
+        Args:
+            id_ (str):
+                As returned by :attr:`Future.id`.
+
+        Returns:
+            :class:`Future`
+        
+        """
+        future = Future(None, id_)
+        self._load(future)
+        return future
+
     def get_solvers(self, refresh=False, order_by='avg_load', **filters):
         """Return a filtered list of solvers handled by this client.
 
@@ -1195,6 +1211,14 @@ class Client(object):
 
                 # If the message is complete, forward it to the future object
                 if 'answer' in message:
+
+                    # If the future does not know which solver it's associated
+                    # with, we get it from the info provided from the server.
+                    # An alternative to making this call here would be to pass
+                    # self in with the message
+                    if future.solver is None:
+                        future.solver = self.get_solver(name=message['solver'])
+
                     future._set_message(message)
                 # If the problem is complete, but we don't have the result data
                 # put the problem in the queue for loading results.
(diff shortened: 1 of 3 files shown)


dwave/cloud/client.py
```diff
@@ -1195,6 +1211,14 @@ def _handle_problem_status(self, message, future):
 
                 # If the message is complete, forward it to the future object
                 if 'answer' in message:
+
+                    # If the future does not know which solver it's associated
+                    # with, we get it from the info provided from the server.
+                    # An alternative to making this call here would be to pass
+                    # self in with the message
+                    if future.solver is None:
```
arcondello: I am not sure how much of a hack it is to do this here. As I mention in the comment, an alternative would be to do something like
```
future._set_message_and_solver(message, client)
```
Thoughts?

arcondello:
Also, I totally missed https://github.com/dwavesystems/dwave-cloud-client/blob/bbbd11772cd6df5295edef611eb6a64dfa62d0be/dwave/cloud/solver.py#L139 before, oops. I can change the method name in this PR to match.

dwave/cloud/client.py
randomir: This whole function is a hack, and I want to refactor it at some point, having in mind we have better control over problem/answer/metadata access on SAPI since recently.

dwave/cloud/client.py
```diff
@@ -646,6 +647,21 @@ def _fetch_solvers(self, name=None):
...
+                As returned by :attr:`Future.id`.
+
+        Returns:
+            :class:`Future`
+        
+        """
+        future = Future(None, id_)
+        self._load(future)
+        return future
+
```
randomir: To be consistent, let's rename this to `retrieve_answer`.

dwave/cloud/client.py
```diff
@@ -646,6 +647,21 @@ def _fetch_solvers(self, name=None):
...
+        Args:
+            id_ (str):
+                As returned by :attr:`Future.id`.
+
+        Returns:
+            :class:`Future`
+        
+        """
+        future = Future(None, id_)
+        self._load(future)
```
randomir: Are you sure this will work when the problem is submitted, it's accepted (we have the ID), but it's still in `PENDING` or `IN_PROGRESS` state?

arcondello: I did check that, but it's been a few days so I will double check when I am making the name change etc

arcondello:
Ok, I ran a bunch of variations on
```
from dwave.cloud import Client

with Client.from_config(solver=dict(qpu=True)) as client:
    solver = client.get_solver()

    f = solver.sample_ising({0: -1}, {}, num_reads=10000)

    pf = None
    ipf = None

    while not f.done():
        if f.remote_status is None:
            pass
        elif f.remote_status == 'PENDING':
            assert f.id is not None
            if pf is None:
                pf = client.retrieve_answer(f.id)
        elif f.remote_status == 'IN_PROGRESS':
            if ipf is None:
                ipf = client.retrieve_answer(f.id)

print(pf.result())
print(ipf.result())
```
which is obviously super hacky but enough to (reasonably) convince me that everything works. Variations included blocking on the retrieved in-and-out of the context, only doing one of them etc etc.

PR #346, opened 2020-01-29
Use the new solver property, `category`

randomir:
With `category` we finally can reliably differentiate between qpu
and software solvers.


randomir:
@JoelPasvolsky, good point. I'll update the docs.

PR #365, opened 2020-02-19
Blocking submission ID getter

randomir:


dwave/cloud/computation.py
```diff
@@ -487,6 +491,43 @@ def cancel(self):
...
+            :exc:`concurrent.futures.TimeoutError`:
+                When `timeout` exceeded, and problem id not ready.
+
+        """
+        if not self._id_ready_event.wait(timeout=timeout):
+            raise TimeoutError("problem id not available yet")
+
+        return self.id
+
+    def set_id(self, id_):
```
arcondello: You could consider doing this in a setter rather than a method. That was there would still be one way to set, but two ways to get (legacy and blocking).

randomir: @arcondello, I like that idea. I just don't like the asymmetry that it brings: `id` setter would be a pair with `get_id` (they both use the `Event`), and then there would be the plain `id` getter...

randomir: Maybe renaming `get_id` to something like `wait_for_id` would help there.

arcondello: I like that, makes it more explicit

PR #373, opened 2020-03-13
Future.sampleset uses SampleSet.from_future

arcondello:


dwave/cloud/computation.py
```diff
@@ -748,23 +747,45 @@ def sampleset(self):
...
 
         sampleset = dimod.SampleSet.from_samples(
             (samples, variables), vartype=vartype,
             energy=self.energies, num_occurrences=self.occurrences,
             info=info, sort_labels=True)
 
-        self._result['sampleset'] = sampleset
+        # this means that samplesets retrieved BEFORE this function are called
+        # are not the same object as after, but it is a simpler implementation
+        self._result['sampleset'] = self._sampleset = sampleset
```
arcondello: The more complicated implementation would be something like
```
try:
    self._result['sampleset'] = self._sampleset
except AttributeError:
    self._result['sampleset'] = self._sampleset = sampleset
```
I didn't go with it because (I think) it's pretty confusing and because doing any reading in the main branch of the try-catch causes an infinite recursion error. But I could do it and make a long comment about it....

randomir: No, I think it's fine as it is.

Issue #353, opened 2020-02-06
Implement `Future.from_problem_id` factory


randomir:
This entails a client refactor to use new SAPI features for downloading problem/answer/meta/context data.

PR #377, opened 2020-04-03
Annotate Future.sampleset with .wait_id method

arcondello:
Part of https://github.com/dwavesystems/dwave-system/issues/284. Enables the workflow
```
sampleset = LeapHybridSampler().sample(bqm)
id_ = sampleset.wait_id()
```

### DIFF
--- a/dwave/cloud/computation.py
+++ b/dwave/cloud/computation.py
@@ -825,6 +825,10 @@ class Future(object):
             energy=self.energies, num_occurrences=self.num_occurrences,
             info=info, sort_labels=True)
 
+        # the id is stored in the info field, but to be consistent with
+        # the samplesets constructed .from_future, we add the method as well
+        sampleset.wait_id = self.wait_id
+
         # this means that samplesets retrieved BEFORE this function are called
         # are not the same object as after, but it is a simpler implementation
         self._result['sampleset'] = self._sampleset = sampleset
@@ -833,7 +837,11 @@ class Future(object):
 
     @property
     def sampleset(self):
-        """Return :class:`~dimod.SampleSet` representation of the results."""
+        """Return :class:`~dimod.SampleSet` representation of the results.
+
+        Adds a `.wait_id` method to retrieve the id before the sampleset is
+        resolved.
+        """
 
         try:
             return self._sampleset
@@ -848,6 +856,9 @@ class Future(object):
 
         self._sampleset = sampleset = dimod.SampleSet.from_future(
             self, lambda f: f.wait_sampleset())
+
+        # propagate id to sampleset as well
+        sampleset.wait_id = self.wait_id
 
         return sampleset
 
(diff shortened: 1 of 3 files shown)


dwave/cloud/computation.py
```diff
@@ -763,6 +763,10 @@ def wait_sampleset(self):
             energy=self.energies, num_occurrences=self.occurrences,
             info=info, sort_labels=True)
 
+        # the id is stored in the info field, but to be consistent with
+        # the samplesets constructed .from_future, we add the method as well
+        sampleset.wait_id = self.wait_id
```
randomir, line 768: I would suggest we use a more specific name on samplesets, e.g. `wait_problem_id`. For future-proofing. It's too late to fix it on computation class (without breaking compat), but I would like to reserve the `id` attr on all Ocean objects for future profiling/tracking/logging purposes.

randomir: OTOH, in line with the [workflow you present](https://github.com/dwavesystems/dwave-cloud-client/pull/377#issue-398363580), I'm thinking something like this would make a nicer interface:
```
sampleset.problem_id(block=True, timeout=None)
```
API inspired by [`queue.SimpleQueue.get`](https://docs.python.org/3/library/queue.html#queue.SimpleQueue.get).

arcondello: Agree. We would add the additional `block` kwarg as part of the existing `Future.wait_id` method? It might be a bit weird with the current method name but probably simpler. Or we could make a new `problem_id` method on `Future` and just alias `wait_id` to it.

randomir: I would prefer to add `Future.problem_id`, or similar (with `block`/`timeout`) and then eventually deprecate the `.id` attr.

We introduced `wait_id`, instead of doing the above in the first place (on `id` method) because `id` existed as an attribute already, and we didn't want to break compat just prior to Leap2 launch.

arcondello: Ok, should be pretty straightforward. I would be happy to do it as part of this PR or I can wait and rebase this one. Let me know.
If the former, just let me know what exception you'd like me to raise in the non-blocking case.

randomir: Can we think about it a bit more? I was thinking about maybe renaming `Future` here as well, and all this would make sense as a part of https://github.com/dwavesystems/dwave-cloud-client/issues/353.

To unblock you, and the work on https://github.com/dwavesystems/dwave-system/issues/284, I think this PR can be merged independently of those other client changes.

randomir: To wrap up this thread.. Until we make a Leap-wide/SAPI-wide problem/job/bqm id naming decision, better not to proceed with `problem_id` approach here for now. Using the ambiguous wait_id, and refactoring later for system-wide consistency might be better.

PR #383, opened 2020-06-05
Add support for client-side certificate auth

randomir:
Config/kwarg options added to client:
- `client_cert`
- `client_cert_key`

They both accept a string path to a crt/key file.

If only `client_cert` is given, the file must contain both the certificate
and the key. Alternatively, cert and key can be split among the two
options.

Closes #383.


charleswhchan:
Thanks! LGTM. You can leave this as a branch and we will test and leave a comment.

No rush to merge this PR until we know it's working e2e.

PR #404, opened 2020-07-21
Fix part upload retry (during multipart upload)

### LINKED ISSUE #403: Upload part worker does not recover from failures to correctly retry upload
Individual part problem upload should be retried 2 times in case the upload fails due to any unhandled exception (e.g. SAPI returns `502 Bad Gateway` during overload). This is achieved with the `retried` decorator.

https://github.com/dwavesystems/dwave-cloud-client/blob/88e14e9b745957569d86ca8d8fcd1461ff7101b9/dwave/cloud/client.py#L1670-L1672

The problem is not all arguments are immutable. `part_stream` (uploaded part data binary stream) should always be at the beginning.

We can either always rewind the `part_stream`, or instead to pass in an immutable "chunk/part stream factory".

randomir:
Closes #403.

### DIFF
--- a/dwave/cloud/client.py
+++ b/dwave/cloud/client.py
@@ -1668,7 +1668,7 @@ class Client(object):
 
     @staticmethod
     @retried(_UPLOAD_PART_RETRIES, backoff=_UPLOAD_RETRIES_BACKOFF)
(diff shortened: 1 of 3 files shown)


randomir:
@oneklc reports he verified the patch by running an extensive load test again.

PR #432, opened 2020-09-23
Add DQM solver

randomir:
- abstract `dwave.cloud.solver.UnstructuredSolver` to `.BaseUnstructuredSolver`
- alias `dwave.cloud.solver.BQMSolver` as `.UnstructuredSolver`
- add `dwave.cloud.solver.DQMSolver`, with `sample_dqm` method, as well as fallback `sample_{bqm,ising,qubo}`

### DIFF
--- a/dwave/cloud/solver.py
+++ b/dwave/cloud/solver.py
@@ -488,6 +488,79 @@ class BQMSolver(BaseUnstructuredSolver):
         return self.upload_problem(bqm)
 
 
+class DQMSolver(BaseUnstructuredSolver):
+    """Class for D-Wave unstructured discrete quadratic model solvers.
+
+    This class provides a :term:`DQM` sampling
+    method and encapsulates the solver description returned from the D-Wave
+    cloud API.
+
+    Args:
+        client (:class:`~dwave.cloud.client.Client`):
+            Client that manages access to this solver.
+
+        data (`dict`):
+            Data from the server describing this solver.
+
+    Note:
+        Events are not yet dispatched from unstructured solvers.
+    """
+
+    _handled_problem_types = {"dqm"}
+    _handled_encoding_formats = {"bq"}
+
+    def _encode_problem_for_upload(self, dqm):
+        try:
+            data = dqm.to_file()
+        except Exception as e:
+            logger.debug("DQM conversion to file failed with %r, "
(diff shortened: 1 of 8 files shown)


dwave/cloud/solver.py
```diff
@@ -408,16 +499,93 @@ def upload_bqm(self, bqm):
...
-            # assume `bqm` given as file, ready for upload
-            data = bqm
+            # assume `dqm` given as file, ready for upload
+            data = dqm
 
-        return self.client.upload_problem_encoded(data)
+        logger.debug("Problem encoded for upload: %r", data)
+        return data
+
+    def sample_bqm(self, bqm, **params):
```
arcondello, line 541: Does this need to be "public"? Just not necessarily something that we want to encourage folks to do

arcondello (approved):

dwave/cloud/solver.py
randomir: Not sure what you mean by "this". `sample_bqm` method?

randomir: Oh, you mean on the `DQMSolver`. Well, to make the new type better fit in, for now, yes.

arcondello: I wonder if maybe a doc string that says "it's better to use a BQM solver" or similar would be good.

randomir: Btw, I would not document this method, that's for sure. Like I wouldn't *many* other "public" methods in this module. /cc @JoelPasvolsky

dwave/cloud/solver.py
```diff
@@ -345,49 +377,108 @@ def _encode_problem_as_ref(self, problem, params):
...
+            solver=self, id_=None, return_matrix=self.return_matrix)
 
         logger.debug("Submitting new problem to: %s", self.id)
         self.client._submit(body, computation)
 
         return computation
 
+
+class BQMSolver(BaseUnstructuredSolver):
+    """Class for D-Wave unstructured binary quadratic model solvers.
```
JoelPasvolsky, line 432: "Class for D-Wave unstructured" --> I think we generally refer to these as Leap rather than D-Wave

JoelPasvolsky (commented):
"handled" is more standardly "supported"

PR #434, opened 2020-10-06
Bubble up SAPI errmsg on upload

### LINKED ISSUE #433: Bubble up SAPI error messages during all stages of multipart upload
For example, if a too big upload size is requested (during multipart upload initialize), SAPI responds with:
```
{"error_code": 400, "error_msg": "The proposed upload is bigger than the maximum upload allowed of 42949672960"}
```
but the exception raised is `HTTPError('400 Client Error: Bad Request ...')`.

The expected behavior in this case is to fail with `ProblemUploadError("The proposed upload is bigger than ...")` or similar.

randomir:
Adds consistent response parsing on all multipart upload endpoints, so that SAPI error message is propagated as `ProblemUploadError` message.

Remaining endpoints (problems/, solvers/) are not yet consistent, as they might not return a JSON error object.

Closes #433.

### DIFF
--- a/dwave/cloud/client.py
+++ b/dwave/cloud/client.py
@@ -44,12 +44,12 @@ import sys
 import time
 import json
 import copy
+import queue
 import logging
-import threading
-import requests
+import inspect
 import warnings
 import operator
-import queue
+import threading
 
 import base64
 import hashlib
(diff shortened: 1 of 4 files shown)


arcondello (commented):
LGTM

PR #444, opened 2021-01-12
Add problem labelling support

randomir:


dwave/cloud/computation.py
```diff
@@ -789,8 +792,8 @@ def wait_sampleset(self):
         vartype_from_problem_type = {'ising': 'SPIN', 'qubo': 'BINARY'}
         vartype = vartype_from_problem_type[self.problem_type]
 
-        # include timing and id in info
-        info = dict(timing=self.timing, problem_id=self.id)
+        # include timing and id/label in info
+        info = dict(timing=self.timing, problem_id=self.id, problem_label=self.label)
```
arcondello, line 796: maybe better to just not include it if it's `None`?

arcondello (approved):
LGTM

JoelPasvolsky:
I'm only partway through but as it's approved I'll ask quickly before it's merge, @randomir can you say something about your considerations on explicitly having label in the sampling functions' parameters as you have versus pulling it out of `**params`? Thanks!

dwave/cloud/computation.py
randomir: Good catch.

randomir:
@JoelPasvolsky, I went with explicit because `label` is not a solver parameter (unlike `**params`). In the past, we added `offset` explicitly as well (which is an optional problem parameter). So, in each `sample_*` method we now actually have 3 categories of arguments: problem description, solver params and common/shared/sampling params (only label for now).

Originally, we thought to have label as a solver parameter, but to allow filtering by label, SSW decided to implement it level up, next to problem id, problem type, status, etc.

Alternatively, we could repurpose `**params` as `**kwargs`, and then split them lower down before the actual submit. Technically, that's trivial, but it's a conceptual change requiring docs update and additional explanation. So, this time I just went along with [#2](https://zen-of-python.info/explicit-is-better-than-implicit.html#2), even if that makes the code/docs more verbose.

JoelPasvolsky:
Thanks, @randomir, that makes sense. I guess the reason this caught my attention is that I feel it's good to have the minimum number of arguments to sampling functions to make those simple conceptually for users, so every explicit argument should carry its weight, and it's not clear to me that most users will often use labels (they're very useful for examples such as those in dwave-examples and animated demos), but it's another argument they need to learn when they're getting started.

dwave/cloud/solver.py
```diff
@@ -306,11 +309,14 @@ def sample_qubo(self, qubo, offset=0, **params):
             offset (optional, default=0):
                 Constant offset applied to the model.
 
+            label (str, optional):
+                Problem label.
```
JoelPasvolsky, line 313: Just wondering whether it's worth the additional length to clarify the optionality of this parameter and its intention as an aid for user identification rather than a system need, for example something like, "Problem label you can optionally tag submissions with for ease of identification."

dwave/cloud/computation.py
randomir: FWIW, in current SAPI implementation, `label` is always returned (set to empty string if undefined), so this won't have effect unless/until SAPI behavior is changed.

arcondello: Cannot say I love that. Feels like if it'll always be an empty string on the way out, then it should default to an empty string on the way in, for symmetry is nothing else. But I gather this touches on a deeper discussion with the SAPI team.

randomir:
@JoelPasvolsky, I agree. FWIW, dimod samplers in dwave-system use kwargs, so we don't have to extract label as explicit argument there. Most users (especially new users) interface only with dwave-system samplers anyhow.

dwave/cloud/computation.py
randomir: Yup, after the latest discussion, they'll do:
- `label` can be omitted or set to `null` on input,
- in which case `label` will be set to `null` on output

PR #471, opened 2021-07-14
Use chardet rather than charset_normalizer

arcondello:
Otherwise multipart upload gets `dwave.cloud.exceptions.ProblemUploadError: Multipart upload of problem_id='...' failed for part_no=1 with InvalidAPIResponseError('JSON response expected').`


hhtong (approved):

arcondello:
See also https://github.com/psf/requests/issues/5871 (just posted a few hours ago).
I imagine there is a more elegant fix, but specifying the encoder explicitly.

PR #474, opened 2021-07-30
Support dimod 0.10

randomir:


.circleci/config.yml
```diff
@@ -28,6 +30,16 @@ jobs:
             pip install -r requirements.txt -r tests/requirements.txt
             pip install wheel twine
 
+      - run:
+          name: Install appropriate dimod version
+          command: |
+            . env/bin/activate
+            if [[ "<< parameters.dimod-version >>" ]]; then
+              pip install 'dimod<< parameters.dimod-version >>'
```
arcondello, line 38: this may mess with whatever numpy version you installed from `requirements.txt`. Probably doesn't matter but just in case.

arcondello (approved):
LGTM!

.circleci/config.yml
randomir: Shouldn't that dimod install pull the appropriate numpy?

arcondello: Yup, just that you already installed one in requirements, so if for whatever reason you wanted to pin that version in the future you might be surprised it gets changed in CI.

randomir: Oh, that's what you mean. Yeah, that's perfectly fine. I use numpy range only here anyway (so it can be determined by dimod).

arcondello: As I said, don't think it matters :shrug:

PR #476, opened 2021-08-10
Don't install requests with the use_chardet_on_py3 extra

spakin:
This extra seems no longer to exist.  I receive
```
WARNING: requests 2.25.1 does not provide the extra 'use_chardet_on_py3'
```
and later
```
Traceback (most recent call last):
  File "/home/pakin/.anaconda3/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2739, in requires
    deps.extend(dm[safe_extra(ext)])
KeyError: 'use_chardet_on_py3'
```
As a result, I was unable to use, in particular, the D-Wave Problem Inspector.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-requests[socks, use_chardet_on_py3]>=2.18
+requests[socks]>=2.18
 homebase>=1.0.0
 python-dateutil>=2.7
 click>=7
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ except SyntaxError:
 python_requires = '>=3.6'
 
 # Package requirements, minimal pinning
-install_requires = ['requests[socks, use_chardet_on_py3]>=2.18',
+install_requires = ['requests[socks]>=2.18',
                     'homebase>=1.0',
                     'click>=7.0', 'python-dateutil>=2.7', 'plucky>=0.4.3']
 


randomir:
https://docs.python-requests.org/en/master/community/updates/#id1

It looks like we should bump the requests lower bound instead.

Also, see #471.

Ousret:
Hi,

I am aware of what caused https://github.com/dwavesystems/dwave-cloud-client/pull/471 and was fixed recently. You should be able to remove Chardet as a dependency. In case there are any concerns left, do not hesitate to reach.

Also, for this matter at hand, don't use the `[use_chardet_on_py3]` but add explicitly `chardet` instead, will do the same without breaking anything for people having `requests` < 2.26.

Regards,

randomir:
Thanks @Ousret for adding more context.

In summary:
- `requests<2.26` uses `chardet` to detect character encoding in a response (even for JSON decoding)
- `requests>=2.26` uses `charset-normalizer~=2.0.0` which fails to provide a meaningful encoding of some JSON strings, like `{}`, and in turn, response `.json()` decoding fails.
- `requests>=2.26` will prefer `chardet` over `charset-normalizer`, if `chardet` is installed.
- You've fixed https://github.com/Ousret/charset_normalizer/issues/58 in `charset-normalizer>=2.0.2`, some 12 days after the 2.0.0 release, and just one day after `requests==2.26` was released.
- A client that expects specific encoding (e.g. UTF-8 for JSON), can specify the encoding explicitly, like the original bug reporter did in https://github.com/googleapis/python-cloud-core/pull/118.
- The complete charset guessing game might stop in the next major `requests` release, as said in https://github.com/psf/requests/issues/5871#issuecomment-879854653

Given all that, we can safely merge this change. Thanks @spakin for bringing it up.

In a follow-up PR I'll explicitly specify the encoding, as that should make response decoding not just resilient to future charset guessing issues, but also faster.

spakin:
Thanks for merging the pull request.  I didn't realize until I read https://github.com/dwavesystems/dwave-cloud-client/pull/476#issuecomment-902314294 what a mess `requests` has produced.

PR #483, opened 2021-08-28
Add multi region support

randomir:


mdecandia:
It appears that specifying a region when instantiating the Client from config does not override the endpoint specified in dwave.conf. For example, if dwave.conf specifies `'https://eu-central-1...` but the client is instantiated as Client.from_config(region="na-west-1"), the client will use the EU endpoint.

dwave/cloud/client/base.py
```diff
@@ -683,6 +737,43 @@ def is_solver_handled(solver):
...
+            metadata_api_endpoint)
+
+        with api.Regions(endpoint=metadata_api_endpoint) as regions:
+            data = regions.list_regions()
+
+        logger.trace("Received region metadata: %r", data)
+
+        return data
+
+    def get_regions(self, refresh=False):
```
randomir, line 753: TODO: make static.

Issue #387, opened 2020-06-29
MockSolver(s)


arcondello:
It would be useful to have mock solvers for the QPU and the leap hybrid samplers for use in testing.

See Also:
https://github.com/dwavesystems/dwave-cloud-client/issues/22 
https://github.com/dwavesystems/dwave-system/issues/283#issuecomment-607360274

PR #486, opened 2021-09-29
Add QPU solvers data mocks

randomir:
Prerequisite for #387.

### DIFF
--- a/dwave/cloud/testing/mocks.py
+++ b/dwave/cloud/testing/mocks.py
@@ -125,11 +125,11 @@ def qpu_clique_solver_data(size: int, **kwargs) -> dict:
     couplers = list(itertools.combinations(range(len(qubits)), 2))
 
     params = dict(
-        id="dw_{}q_mock".format(size),
-        description="A {}-qubit clique mock QPU solver".format(size),
+        id=f"clique_{size}q_mock",
+        description=f"A {size}-qubit mock QPU solver with clique topology",
         qubits=qubits,
         couplers=couplers,
-        topology={"type": "clique"},
+        topology={"type": "clique", "shape": [size]},
     )
     params.update(**kwargs)
 
@@ -144,7 +144,8 @@ def qpu_chimera_solver_data(m: int,
 
     Args:
         m:
-            Number of rows in the Chimera lattice.
+            Number of rows in the Chimera lattice. See
+            :func:`~dwave_networkx.generators.chimera_graph` for details.
         n:
             Number of columns in the Chimera lattice.
         t:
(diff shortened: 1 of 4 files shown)


dwave/cloud/testing/mocks.py
```diff
@@ -135,6 +136,78 @@ def qpu_clique_solver_data(size: int, **kwargs) -> dict:
...
+    qubits = list(graph.nodes)
+    couplers = list(graph.edges)
+    num_qubits = len(qubits)
+
+    params = dict(
+        id=f"dw_{num_qubits}q_mock",
+        description=f"A {num_qubits}-qubit mock QPU solver with chimera topology",
+        qubits=qubits,
+        couplers=couplers,
+        topology={"type": "chimera", "shape": [m, n, t]},
```
JoelPasvolsky, line 171: How is the topology propagated through `structured_solver_data`?

dwave/cloud/testing/mocks.py
```diff
@@ -135,6 +136,78 @@ def qpu_clique_solver_data(size: int, **kwargs) -> dict:
...
+        raise RuntimeError("Can't generate Pegasus graph without dwave-networkx. "
+                           "Install with 'dwave-cloud-client[mocks]'.")
+
+    graph = dnx.pegasus_graph(m, fabric_only=False)
+    qubits = list(graph.nodes)
+    couplers = list(graph.edges)
+    num_qubits = len(qubits)
+
+    params = dict(
+        id=f"dw_{num_qubits}q_mock",
```
JoelPasvolsky, line 200: Should this be `f"Advantage_{num_qubits}q_mock" ?

requirements.txt
```diff
@@ -13,5 +13,8 @@ dataclasses; python_version == "3.6"
 dimod>=0.10.5
 numpy>=1.16
 
+# for solver mocks
+dwave-networkx>=0.8.9
```
JoelPasvolsky, line 17: Does this new dependency deserve a comment in reno?

dwave/cloud/testing/mocks.py
randomir: It's just a default name, easily overridden with `id` kwarg, but sure, we can make it better.

dwave/cloud/testing/mocks.py
```diff
@@ -135,6 +136,78 @@ def qpu_clique_solver_data(size: int, **kwargs) -> dict:
...
+        **kwargs:
+            Solver properties passed down to :meth:`.structured_solver_data`.
+    """
+    try:
+        import dwave_networkx as dnx
+    except ImportError:     # pragma: no cover
+        raise RuntimeError("Can't generate Pegasus graph without dwave-networkx. "
+                           "Install with 'dwave-cloud-client[mocks]'.")
+
+    graph = dnx.pegasus_graph(m, fabric_only=False)
```
JoelPasvolsky, line 194: Why `fabric_only=False` here? Online Advantage QPUs usually have qubits from 30 and up

JoelPasvolsky (approved):
LGTM

requirements.txt
randomir: I don't think it does. It's just an optional dependency (like dimod). And we use it just for testing.

dwave/cloud/testing/mocks.py
randomir: Good catch. I had it like this initially to match the `num_qubits`, but it makes sense to be closer to an actual online qpu.

requirements.txt
JoelPasvolsky: We say that to install from source, users should do `pip install -r requirements.txt` and that now has `dwave-networkx` (as well as dimod). How would users know it's optional?

randomir: Because it's in package's `mocks` extra. Like dimod is in `bqm`/`dqm` extra. (In setup.py, and declared in package metadata).

JoelPasvolsky: Perfect, thanks!

PR #488, opened 2021-10-01
Improve check_problem

### LINKED ISSUE #487: Optimize `StructuredSolver.check_problem()`
Currently problem graph check is naively implemented and relatively slow.
```
import dimod
from dwave.cloud.solver import StructuredSolver
from dwave.cloud.testing import mocks

C16 = StructuredSolver(data=mocks.qpu_chimera_solver_data(16), client=None)
P16 = StructuredSolver(data=mocks.qpu_pegasus_solver_data(16), client=None)

bqm_c = dimod.generators.ran_r(1, (C16.nodes, C16.edges))
bqm_p = dimod.generators.ran_r(1, (P16.nodes, P16.edges))
```
```
In [1]: %timeit C16.check_problem(bqm_c.linear, bqm_c.quadratic)
5.04 ms ± 89.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [2]: %timeit P16.check_problem(bqm_p.linear, bqm_p.quadratic)
36 ms ± 871 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

randomir:
Implements 15-20% speed-up of `StructuredSolver.check_problem()` and a new exception that `DWaveSampler` can use to skip (the duplicated) problem check.

Close #487.

### DIFF
--- a/dwave/cloud/solver.py
+++ b/dwave/cloud/solver.py
@@ -31,7 +31,8 @@ You can list all solvers available to a :class:`~dwave.cloud.client.Client` with
 import json
 import logging
 import warnings
-from collections import abc
+from collections.abc import Iterable, Mapping
+from typing import Tuple
 
 from dwave.cloud.exceptions import *
 from dwave.cloud.coders import (
@@ -955,7 +956,7 @@ class StructuredSolver(BaseSolver):
             # that they match because lin can be either a list or a dict. In the future it would be
             # good to check.
             initial_state = params['initial_state']
-            if isinstance(initial_state, abc.Mapping):
+            if isinstance(initial_state, Mapping):
 
                 initial_state_list = [3]*self.properties['num_qubits']
 
@@ -983,7 +984,7 @@ class StructuredSolver(BaseSolver):
                 Quadratic terms of the model (J).
 
         Returns:
-            boolean
+            bool
 
         Examples:
             This example creates a client using the local system's default D-Wave Cloud Client
@@ -1005,13 +1006,11 @@ class StructuredSolver(BaseSolver):
             False
             True
         """
-        for key, value in uniform_iterator(linear):
-            if value != 0 and key not in self.nodes:
-                return False
-        for key, value in uniform_iterator(quadratic):
-            if value != 0 and tuple(key) not in self.edges:
-                return False
-        return True
+        # handle legacy format
+        if not isinstance(linear, Mapping):
+             linear = {idx: val for idx, val in enumerate(linear) if val != 0}
+
+        return self.nodes.issuperset(linear) and self.edges.issuperset(quadratic)
 
 
 # for backwards compatibility:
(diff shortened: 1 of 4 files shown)


dwave/cloud/solver.py
```diff
@@ -1005,13 +1008,11 @@ def check_problem(self, linear, quadratic):
...
-        for key, value in uniform_iterator(linear):
-            if value != 0 and key not in self.nodes:
-                return False
-        for key, value in uniform_iterator(quadratic):
-            if value != 0 and tuple(key) not in self.edges:
-                return False
-        return True
+        # handle legacy format
+        if not isinstance(linear, Mapping):
+             linear = {idx: val for idx, val in enumerate(linear) if val != 0}
```
arcondello, line 1013: It doesn't really matter, but since you're just checking the keys of `linear` you could do
```python
linear = [idx for idx, val in enumerate(linear) if val != 0]
```
or even use [np.nonzero](https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html).

arcondello (approved):
LGTM!

dwave/cloud/solver.py
randomir: I went with a dict because it's already set-like for keys, and `issuperset(linear)` doesn't have to build a set. Not sure it actually make any difference, especially for a small list of linear terms...

arcondello: Ah, interesting! It does indeed seem like they are casting to a set under the hood (at least based on timings)
```python
In [1]: a = set(range(1000))

In [2]: b_list = list(range(500, 1000))

In [3]: b_set = set(range(500, 1000))

In [4]: b_dict = {v: 0 for v in range(500, 1000)}

In [5]: %timeit a.issuperset(b_list)
13.2 µs ± 68.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [6]: %timeit a.issuperset(set(b_list))
13.4 µs ± 183 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [7]: %timeit a.issuperset(b_set)
7.25 µs ± 21.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [8]: %timeit a.issuperset(b_dict)
11.6 µs ± 48.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [9]: %timeit a.issuperset(set(b_dict))
11.8 µs ± 69.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
which kind of surprises me. I would have expected that any iterable would be fine.

arcondello: And so they do https://github.com/python/cpython/blob/f25f2e2e8c8e48490d22b0cdf67f575608701f6f/Objects/setobject.c#L1740

PR #489, opened 2021-10-04
Add CQM solver support

randomir:


dwave/cloud/solver.py
```diff
@@ -617,6 +619,98 @@ def upload_dqm(self, dqm):
...
+        try:
+            import dimod
+        except ImportError: # pragma: no cover
+            raise RuntimeError("Can't sample from 'cqm' without dimod. "
+                               "Re-install the library with 'cqm' support.")
+
+        # to sample BQM problems, we need to convert them to CQM
+        cqm = dimod.CQM.from_bqm(bqm)
+
+        # TODO: convert sampleset back
```
arcondello, line 665: in the BQM case there is no conversion back from sample set. For DQM you do need to do some reinterpretation.

setup.py
```diff
@@ -32,6 +32,9 @@
     # dqm support
     'dqm': ['dimod>=0.9.7,<0.11,!=0.10.0,!=0.10.1,!=0.10.2,!=0.10.3,!=0.10.4', 'numpy>=1.16'],
 
+    # cqm support
+    'dqm': ['dimod>=0.10.5,<0.11', 'numpy>=1.16'],
```
arcondello, line 36: `'cqm'`

arcondello (approved):

dwave/cloud/solver.py
arcondello: Hmm, interesting question. IMO since at the moment we don't enforce the feasibility fields by subclass, having "extra" data vectors is totally fine. In the future we will probably have a `ConstrainedSampleSet` or something like that at which point we'd want to do some conversion.

PR #500, opened 2021-10-31
Add license file to distribution

BastianZim:
Currently, the license file is not included in the PyPI distribution. This should fix it.

Ref: https://pypi.org/project/dwave-cloud-client/#files

CC https://github.com/dwavesystems/dwave-ocean-sdk/issues/144

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,6 +62,7 @@ setup(
     long_description=open('README.rst', encoding='utf-8').read(),
     url=package_info['__url__'],
     license=package_info['__license__'],
+    license_files=["LICENSE"],
     packages=packages,
     python_requires=python_requires,
     install_requires=install_requires,


BastianZim:
@randomir This would also make https://github.com/dwavesystems/dwave-ocean-sdk/issues/144 slightly easier but not a requirement.

setup.py
```diff
@@ -62,6 +62,7 @@
     long_description=open('README.rst', encoding='utf-8').read(),
     url=package_info['__url__'],
     license=package_info['__license__'],
+    license_files=["LICENSE"],
```
randomir, line 65: It's interesting [setuptools docs](https://setuptools.pypa.io/en/latest/references/keywords.html) say:
> [...] If neither `license_file` nor `license_files` is specified, this option defaults to `LICEN[CS]E*`, [...]

Could be our license file is not included due to older setuptools used.

Anyway, good to be explicit (as we are in most Ocean packages).

randomir (approved):
LGTM.

setup.py
BastianZim: To be honest, outside of including a license in a pyproject.toml or manifest.in, including licenses has always been flaky for me, apparently setuptools doesn't like them.

BastianZim:
Thanks for merging.

setup.py
randomir: We usually include it in `setup.cfg`, like [here](https://github.com/dwavesystems/dwave-greedy/blob/master/setup.cfg). That seems to work ok. And it amounts to the same thing you did in this PR.

BastianZim: True, that works well too. 
Let me know if you want me to change it.

randomir: No, this is fine. It's consistent with this repo setup config. We can move to setup.cfg on a bigger refactor.

PR #501, opened 2021-11-02
Filter out secrets from logs

### LINKED ISSUE #391: Scrub tokens from logs

randomir:
Close #391.

### DIFF
--- a/dwave/cloud/__init__.py
+++ b/dwave/cloud/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import re
 import sys
 import logging
 import importlib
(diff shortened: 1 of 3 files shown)


hhtong (approved):
LGTM

JoelPasvolsky (approved):
My only question is whether for some case the token prefix itself might contain information that we might want to redact, for example if a customer has multiple projects and doesn't want to expose the relative allocation and frequency of usage of some of those.

randomir:
@JoelPasvolsky, that's a good point. However, omitting the prefix would significantly reduce the usefulness of redacted token.

Issue #465, opened 2021-04-29
Expose reformatted parameters


arcondello:
The [`StructuredSolver._format_params`](
https://github.com/dwavesystems/dwave-cloud-client/blob/b008a3276bdfa091584563a591c5eb8f0f0f94f2/dwave/cloud/solver.py#L950) method currently transforms the `initial_state` parameter from a dict to the format expected by the endpoint.

This translation, while useful for the user, makes it hard for other libraries (e.g. [aws-braket-ocean-plugin](https://github.com/aws/amazon-braket-ocean-plugin-python)) that are trying to mock the behavior of the `DWaveSampler` from [dwave-system](https://github.com/dwavesystems/dwave-system) to do so.

It would be helpful to expose this logic in a way that can be used by other libraries. Some possible approaches:
1) Change the `_format_params` method to a public class method, or even a static method. This would allow users/libaries to use the method without needing to authenticate with D-Wave.
2) Move `_format_parmas` to the `DWaveSampler` level, and expose it as a public method. We would likely want to deprecate the method in the cloud-client.

I am inclined towards (2).

JoelPasvolsky:
2 makes sense to me given that the transformation is specific to QPU. Where would it live in 1? Also I think with 1 it would need renaming, but that may be true for 2 also

randomir:
Keeping this particular function in the cloud client makes sense to me. Why? Because it translates a nice state map into a list with 3's - very SAPI-specific, low-level and ugly.

This is related to #166 and #199.

I'd say, ideally, we don't use SAPI-specific data encoding anywhere on the client's API surface, and the translation happens internally.

PR #510, opened 2022-01-24
Add `StructuredSampler.reformat_parameters()` method

arcondello:
Closes https://github.com/dwavesystems/dwave-cloud-client/issues/465

### DIFF
--- a/dwave/cloud/solver.py
+++ b/dwave/cloud/solver.py
@@ -28,8 +28,10 @@ You can list all solvers available to a :class:`~dwave.cloud.client.Client` with
 
 """
 
+import copy
 import json
 import logging
+import typing
 import warnings
 from collections.abc import Mapping
 
@@ -51,6 +53,11 @@ try:
 except ImportError:
     _numpy = False
 
+try:
+    import dimod
+except ImportError:
+    dimod = None
+
 __all__ = [
     'BaseSolver', 'StructuredSolver',
     'BaseUnstructuredSolver', 'UnstructuredSolver',
@@ -58,6 +65,19 @@ __all__ = [
 ]
 
 logger = logging.getLogger(__name__)
+
+
+if typing.TYPE_CHECKING:
+    # do a bit of fiddling
+    try:
+        _Type = typing.Literal['qubo', 'ising']
+    except ImportError:  # Python < 3.8
+        _Type = str
+
+    try:
+        _Vartype = typing.Union[_Type, dimod.typing.VartypeLike]
+    except AttributeError:  # dimod not installed or too old
+        _Vartype = _Type
 
 
 class BaseSolver(object):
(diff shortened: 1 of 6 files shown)


dwave/cloud/solver.py
```diff
@@ -1043,19 +1063,62 @@ def _sample(self, type_, linear, quadratic, offset, params,
...
+        Currently the only reformatted parameter is ``initial_state``. This
+        method allows ``initial_state`` to be submitted as a dictionary
+        mapping the qubits to their initial value.
 
-                low = -1 if type_ == 'ising' else 0
+        Args:
+            vartype: One of ``'ising'`` or ``'qubo'``. If :mod:`dimod` is
+                installed, this can also be any
+                :class:`~dimod.typing.VartypeLike`.
+            parameters: The parameters to submit to ths solver.
```
JoelPasvolsky, line 1088: ths

dwave/cloud/solver.py
```diff
@@ -1043,19 +1063,62 @@ def _sample(self, type_, linear, quadratic, offset, params,
...
+            properties: The solver's properties. Note that this will
+                work with either :attr:`StructuredSolver.properties`
+                or :attr:`dwave.systems.DWaveSampler.properties`.
```
JoelPasvolsky, line 1091:
```suggestion
            properties: The solver's properties. Supported values 
                are :attr:`StructuredSolver.properties`,
                :attr:`dwave.systems.DWaveSampler.properties`,
                or similar.
```

dwave/cloud/solver.py
```diff
@@ -1043,19 +1063,62 @@ def _sample(self, type_, linear, quadratic, offset, params,
...
+
+        Returns:
+            The reformatted solver parameters.
+            If ``inplace`` this will be the ``parameters``, modified.
+            If ``not inplace`` then this will be a deep copy of ``parameters``,
+            with the relevant fields updated.
```
JoelPasvolsky, line 1097:
```suggestion
            If ``inplace`` is True, this is the modified ``parameters``.
            If ``inplace`` is False, this is a deep copy of ``parameters``
            with the relevant fields updated.
```

dwave/cloud/solver.py
arcondello: I don't think that's actually very clear. I could say ":attr:`StructuredSolver.properties` ... can be provided."

arcondello:
I don't think the current failures are from me.

dwave/cloud/solver.py
JoelPasvolsky: In code, yes, but here's `inplace` is a Boolean argument of the function, so we always use the "arg is True" & "arg is False" format (and never `not arg`).

arcondello: Ok, but why?

dwave/cloud/solver.py
JoelPasvolsky: There's an ambiguity with the "will work with either": do you mean it has to be one of those or that it can be one of those. I understood the latter

arcondello: It can be either, or other options as well.

JoelPasvolsky: How about that (edited)?

dwave/cloud/solver.py
JoelPasvolsky: So users can parse English and not Python: "if <varname> is <value>" is a grammatical clause, while "if <varname>" is a syntactically correct Python structure

dwave/cloud/solver.py
arcondello: My concern is more with "can be set to" as applied to an attribute, a mutable one at that. I will perhaps just drop the clause about the properties altogether and move that to an examples section.

dwave/cloud/solver.py
JoelPasvolsky: Yup, but here you're saying `inplace` (variable name) and not "in place" (preposition)

arcondello: Ok, I concede. And I see that they use this convention in the Python docs as well.

arcondello: Updated.

dwave/cloud/solver.py
arcondello: Updated

randomir (approved):
LGTM!

G as in great. :wink:

PR #525, opened 2022-05-25
Fix config update with mutually exclusive options on the same level

### LINKED ISSUE #524: Mutually exclusive (endpoint, region) options not correctly applied
In 0.10.0, CLI commands like `dwave solvers` do not properly prioritize `endpoint` config option when provided in combination with `region` (on the same config level).

randomir:
Close #524.


arcondello (approved):

randomir:
Rebased to add reno and combine test commits.

PR #530, opened 2022-08-08
Add ``estimate_qpu_access_time`` method

JoelPasvolsky:
Alternative to https://github.com/dwavesystems/dwave-system/pull/461. If we want it in this repo, I'll add docs and a nicer example.


dwave/cloud/utils.py
```diff
@@ -801,3 +802,197 @@ def bqm_to_dqm(bqm):
         dqm.set_quadratic(u, v, biases)
 
     return dqm
+
+def estimate_qpu_access_time(solver,
```
randomir, line 806: Let's make this a method on [`StructuredSolver`](https://github.com/dwavesystems/dwave-cloud-client/blob/442211d5279e9e6a96dc42175549caef66844970/dwave/cloud/solver.py#L732). It'll be in a good company there with similar methods like: `max_num_reads()`, `reformat_parameters()`, `check_problem()`.

`solver` becomes `self`.

dwave/cloud/utils.py
```diff
@@ -801,3 +802,197 @@ def bqm_to_dqm(bqm):
...
+        raise ValueError("set only one of ``anneal_schedule`` or ``annealing_time``")
+
+    if anneal_schedule and anneal_schedule[0][1] == 1 and not initial_state:
+        raise ValueError("reverse annealing requested (``anneal_schedule`` "
+                         "starts with s==1) but ``initial_state`` not provided")
+
+    try:
+        problem_timing_data = solver.properties['problem_timing_data']
+
+        version_timing_model = problem_timing_data['version']
```
randomir, line 907: I suggest a slight re-org here:
- split this big try/catch in three
- in first, test for `problem_timing_data` (or fail appropriately) -> no ugly switch in except block
- in second, test version (or fail appropriately)
- in third, read out the required properties (or fail)

dwave/cloud/utils.py
```diff
@@ -801,3 +802,197 @@ def bqm_to_dqm(bqm):
...
+        elif err == 'version':
+            err_msg = f"selected solver is missing ``problem_timing_data`` field {err}"
+        else:
+            err_msg = f"selected solver is missing ``problem_timing_data`` field " + \
+                      f"{err} required by timing model version {version_timing_model}"
+        raise ValueError(err_msg)
+
+    # Support for sapi timing model versions: 1.0.x
+    version_tuple = tuple(int(i) for i in version_timing_model.split("."))
+    if not version_tuple[0] == 1 and not version_tuple[1] == 0:
```
randomir, line 934: Keep it simple:
```suggestion
    if not version_timing_model.startswith("1.0."):
```

dwave/cloud/utils.py
```diff
@@ -801,3 +802,197 @@ def bqm_to_dqm(bqm):
...
+
+    ra_programming_time = 0
+    ra_delay_time = 0
+    if anneal_schedule and anneal_schedule[0][1] == 1:
+        if reinitialize_state:
+            ra_programming_time = ra_with_reinit_prog_time_delta
+            ra_delay_time = ra_with_reinit_delay_time_delta
+        else:
+            ra_programming_time = ra_without_reinit_prog_time_delta
+            ra_delay_time = ra_without_reinit_delay_time_delta
```
randomir, line 947: Replace `=` with `+=` for clarity and future-proofing (moving code sections around).

dwave/cloud/utils.py
```diff
@@ -801,3 +802,197 @@ def bqm_to_dqm(bqm):
...
+    else:
+        programming_thermalization_time = default_programming_thermalization
+
+    programming_time = typical_programming_time + ra_programming_time + \
+                       programming_thermalization_time
+
+    anneal_time = default_annealing_time
+    if annealing_time:
+        anneal_time = annealing_time
+        print(anneal_time)
```
randomir, line 960: Astray print.

dwave/cloud/utils.py
```diff
@@ -801,3 +802,197 @@ def bqm_to_dqm(bqm):
...
+
+    n = len(readout_time_model_parameters)
+    if n % 2:
+         raise ValueError(f"for the selected solver, ``problem_timing_data`` "
+                          f"property field ``readout_time_model_parameters`` "
+                          f"is not of an even length as required by timing "
+                          f"model version {version_timing_model}")
+
+    q = readout_time_model_parameters[:int(n/2)]
+    t = readout_time_model_parameters[int(n/2):]
```
randomir, line 972: For positive `n`, simpler to use floor division:
```suggestion
    q = readout_time_model_parameters[:n//2]
    t = readout_time_model_parameters[n//2:]
```

dwave/cloud/utils.py
JoelPasvolsky: Can you clarify? These are not accumulating, they're being set to the property value versus (currently) being zeroed; if in the future the potential for RA always adds a small constant bias, replacing = with += will instead introduce an error.

randomir: Ok, in that case, it's fine like this. It's just when I see 2 variables initialized in 3 places, I want to refactor -- to initialize only once, and then update as needed. Same effect here, though.

dwave/cloud/solver.py
```diff
@@ -1174,6 +1174,212 @@ def check_problem(self, linear, quadratic):
...
+            readout_thermalization:
+                Set to ``True`` if your submission sets ``readout_thermalization``.
+            reduce_intersample_correlation:
+                Set to ``True`` if your submission sets ``reduce_intersample_correlation``.
+
+        Returns:
+            Estimated QPU access time, in microseconds.
+
+        Raises:
+            KeyError: If a solver property, or a field in the ``problem_timing_data``
```
randomir, line 1246: I don't think `KeyError` is raised anymore.

dwave/cloud/solver.py
```diff
@@ -1174,6 +1174,212 @@ def check_problem(self, linear, quadratic):
...
+        the same tool on your problem to estimate the expected number of qubits:
+        for large, complex problems you might run the tool several times and take
+        the number of qubits from the produced average or worst-case embedding;
+        for small, simple problems even a single run might be sufficient. If you
+        are using such a tool directly (e.g., in conjunction with the
+        :class:`~dwave.system.composites.FixedEmbeddingComposite`)
+        or otherwise generating a heuristic or non-heuristic embedding, take
+        the required number of qubits from your embedding. Because embedding
+        depends on a QPU’s working graph, such embeddings should be for
+        the particular QPU for which you are estimating the access time.
```
randomir, line 1214: Given this method is in cloud-client now, where all problems are already embedded, perhaps we can simplify the embedding discussion?

tests/test_solver.py
```diff
@@ -120,6 +120,41 @@ def _submit_and_check(self, solver, linear, quad, **kwargs):
...
+    """Test solver utility methods."""
+
+    def test_qpu_access_time_estimate(self):
+        with Client(**config) as client:
+            solver = client.get_solver(qpu=True)
+
+            runtime_q1000 = solver.estimate_qpu_access_time(num_qubits=1000)
+            # results for num_qubits=1000: Advanatge4.1 15341, DW_2000Q_6 11775
+            self.assertGreater(runtime_q1000, 5000)
+            self.assertGreater(50000, runtime_q1000)
```
randomir, line 134: How about you verify the estimate by comparing it with actual access time (since this is a live test, you might as well sample)?

JoelPasvolsky: That's tempting but the estimate varies by selected system, and available systems can change over time, and there can be significant variations (especially between chips with different topologies, but also the same topology), so that's why I set a large range.

tests/test_solver.py
```diff
@@ -120,6 +120,41 @@ def _submit_and_check(self, solver, linear, quad, **kwargs):
 
         return results
 
+@unittest.skipUnless(config, "No live server configuration available.")
+class Utilities(_QueryTest):
+    """Test solver utility methods."""
+
+    def test_qpu_access_time_estimate(self):
```
randomir, line 127: It would be great to run these tests on a mock solver, created from e.g. [`qpu_pegasus_solver_data`](https://github.com/dwavesystems/dwave-cloud-client/blob/442211d5279e9e6a96dc42175549caef66844970/dwave/cloud/testing/mocks.py#L183) -- where you can control estimation params. See [here](https://github.com/dwavesystems/dwave-cloud-client/blob/master/tests/test_solver_utils.py) for an example.

dwave/cloud/solver.py
JoelPasvolsky: Copy & paste error when I split the ``try:`` statement. Fixed

tests/test_solver.py
randomir: If you feel like going down that route (using mock solver for testing the estimator), I would still leave a single/smoke test on a live solver (end-to-end test + SAPI test).

tests/test_solver.py
JoelPasvolsky: Added sampling versus estimate unit test

tests/test_solver.py
JoelPasvolsky: Added mock for qpu_acess_time estimates

dwave/cloud/testing/mocks.py
```diff
@@ -222,6 +222,70 @@ def qpu_pegasus_solver_data(m: int,
...
+        'default_annealing_time': 20.0,
+        'readout_time_model': 'pwl_log_log',
+        'readout_time_model_parameters': [0.0, 3.30984300471607, 2.2975416678181597,
+            2.2975416678181597],
+        'qpu_delay_time_per_sample': 20.54,
+        'reverse_annealing_with_reinit_delay_time_delta': 586.38,
+        'reverse_annealing_without_reinit_delay_time_delta': -5.0,
+        'default_readout_thermalization': 0.0,
+        'decorrelation_max_nominal_anneal_time': 2000.0,
+        'decorrelation_time_range': [500.0, 10000.0]}
```
randomir, line 268: Let's extract these two in a helper function `qpu_problem_timing_data` (get rid of `qpu_access_time_data`), and then simply use that function from the unittests below as:
```
C16 = StructuredSolver(data=mocks.qpu_chimera_solver_data(16, 
    problem_timing_data=qpu_problem_timing_data(generation='2000q')), client=None)
```
Or split `qpu_problem_timing_data` in `qpu_chimera_problem_timing_data` and `qpu_pegasus_problem_timing_data`?

dwave/cloud/testing/mocks.py
```diff
@@ -222,6 +222,70 @@ def qpu_pegasus_solver_data(m: int,
...
+    if qpu == 'Advantage_system4.1':
+        problem_timing_data = timing_data_advantage41_2022_8
+    elif qpu == 'DW_2000Q_6':
+        problem_timing_data = timing_data_2000q6_2022_8
+    elif qpu == 'version_1_1_0':
+        problem_timing_data = timing_data_advantage41_2022_8
+        problem_timing_data["version"] = '1.1.0'
+    elif qpu == 'no_version':
+        problem_timing_data = timing_data_advantage41_2022_8
+        del problem_timing_data['version']
```
randomir, line 279: These trivial modifications belong to a unittest.

tests/test_solver.py
```diff
@@ -121,40 +124,66 @@ def _submit_and_check(self, solver, linear, quad, **kwargs):
...
-class Utilities(_QueryTest):
+class Utilities(unittest.TestCase):
     """Test solver utility methods."""
 
-    def test_qpu_access_time_estimate(self):
-        with Client(**config) as client:
-            solver = client.get_solver(qpu=True)
-
-            runtime_q1000 = solver.estimate_qpu_access_time(num_qubits=1000)
+    def test_qpu_access_time_estimate_mock(self):
```
randomir, line 130: This test should not be guarded by `skipUnless` from above. We want to run as many tests we can if offline.

tests/test_solver.py
```diff
@@ -121,40 +124,66 @@ def _submit_and_check(self, solver, linear, quad, **kwargs):
...
+        solver_mock = StructuredSolver(data=mocks.qpu_access_time_data(), client=None)
+
+        # not allowed annealing_time together with anneal_schedule
+        with self.assertRaises(ValueError):
+            solver_mock.estimate_qpu_access_time(num_qubits=1000,
+                                            anneal_schedule=[[0.0, 0.0], [50.0, 0.5]],
+                                            annealing_time=150)
```
randomir, line 137: You can split subtests into separate tests (initialize `solver_mock` in `setUp`/`setUpClass`), or use [`subTest`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest). I'd vote for former.

tests/test_solver.py
```diff
@@ -121,40 +124,66 @@ def _submit_and_check(self, solver, linear, quad, **kwargs):
         return results
 
 @unittest.skipUnless(config, "No live server configuration available.")
-class Utilities(_QueryTest):
+class Utilities(unittest.TestCase):
```
randomir, line 127: Just to group similar tests, I'd put these in `tests/test_solver_utils.py`.

tests/test_solver.py
JoelPasvolsky: Thanks, was a leftover from when all the tests were together

dwave/cloud/testing/mocks.py
JoelPasvolsky: Implemented first option

tests/test_solver_utils.py
```diff
@@ -113,3 +116,68 @@ def test_vartype_dimod_smoke(self):
...
+
+        # not allowed annealing_time together with anneal_schedule
+        with self.assertRaises(ValueError):
+            solver_mock.estimate_qpu_access_time(num_qubits=1000,
+                                            anneal_schedule=[[0.0, 0.0], [50.0, 0.5]],
+                                            annealing_time=150)
+        # reverse anneal without required initial state
+        with self.assertRaises(ValueError):
+            solver_mock.estimate_qpu_access_time(num_qubits=1000,
+                                            anneal_schedule=[[0.0, 1.0], [2.75, 0.45], [82.75, 0.45], [82.761, 1.0]])
```
randomir, line 135: nit: non pep8 indent

tests/test_solver_utils.py
```diff
@@ -113,3 +116,68 @@ def test_vartype_dimod_smoke(self):
...
+
+        # currently support is for version 1.0.x
+        with self.assertRaises(ValueError):
+            solver_mock_ver = copy.deepcopy(solver_mock)
```
randomir, line 139: This is fine, but just FYI, [`with mock.patch(...)`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) would fit in nicely here.

tests/test_solver_utils.py
```diff
@@ -113,3 +116,68 @@ def test_vartype_dimod_smoke(self):
...
+            solver_mock_ver.properties["problem_timing_data"].pop("version")
+            solver_mock_ver.estimate_qpu_access_time(num_qubits=1000)
+
+        for d, t in [({'num_qubits': 1000}, 15341),
+                     ({'num_qubits': 1000, 'num_reads': 500}, 149225),
+                     ({'num_qubits': 1000, 'annealing_time': 400}, 15721),
+                     ({'num_qubits': 1000, 'anneal_schedule': [[0.0, 0.0], [800.0, 0.5]]}, 16121),
+                     ({'num_qubits': 1000, 'anneal_schedule': [[0.0, 1.0], [2, 0.45], [102, 0.45], [102, 1.0]], 'initial_state': {node: 0 for node in solver_mock.nodes}}, 15427),
+                     ]:
+            self.assertEqual(int(solver_mock.estimate_qpu_access_time(**d)), t)
```
randomir, line 154: Good fit for `parameterized` (e.g. see [here](https://github.com/dwavesystems/dwave-cloud-client/blob/442211d5279e9e6a96dc42175549caef66844970/tests/test_mock_submission.py#L1004)).

tests/test_solver.py
JoelPasvolsky: Implemented with ``setUpClass``

tests/test_solver.py
```diff
@@ -37,6 +37,9 @@
     CanceledFutureError, SolverFailureError, InvalidProblemError)
 import dwave.cloud.computation
 
+from dwave.cloud.solver import StructuredSolver
+from dwave.cloud.testing import mocks
```
randomir, line 41: Unused.

tests/test_solver_utils.py
```diff
@@ -113,3 +116,76 @@ def test_vartype_dimod_smoke(self):
...
+class QpuAccessTimeEstimate(unittest.TestCase):
+    """Test QPU access time estimation method."""
+
+    @classmethod
+    def setUpClass(cls):
+        if not dimod:
+            raise unittest.TestCase.skipTest(cls, "dimod unavailable so no dwave_networkx required for mock")
+
+        cls.solver_mock = StructuredSolver(data=mocks.qpu_pegasus_solver_data(16,
+            problem_timing_data=mocks.qpu_problem_timing_data(qpu='advantage')), client=None)
```
randomir, line 129:
```suggestion
        try:
            cls.solver_mock = StructuredSolver(data=mocks.qpu_pegasus_solver_data(16,
                problem_timing_data=mocks.qpu_problem_timing_data(qpu='advantage')), client=None)
        except RuntimeError:
            raise cls.skipTest("structured solver mock not available")
```

randomir (approved):
LGTM!

PR #540, opened 2023-01-06
Add LeapAPI support

randomir:
In this PR we add `LeapAPIClient`  base client, `LeapAccount` resource class, and the appropriate Leap API models (types) in order to support:
- fetching the active project in Leap
- listing all user projects
- fetching SAPI token for each of the available projects

### DIFF
--- a/dwave/cloud/api/client.py
+++ b/dwave/cloud/api/client.py
@@ -26,7 +26,7 @@ from dwave.cloud.package_info import __packagename__, __version__
 from dwave.cloud.utils import (
     TimeoutingHTTPAdapter, BaseUrlSession, user_agent, is_caused_by)
 
-__all__ = ['DWaveAPIClient', 'SolverAPIClient', 'MetadataAPIClient']
+__all__ = ['DWaveAPIClient', 'SolverAPIClient', 'MetadataAPIClient', 'LeapAPIClient']
 
 logger = logging.getLogger(__name__)
 
@@ -327,6 +327,13 @@ class DWaveAPIClient:
         return retry
 
     @classmethod
+    def _add_auth(cls, session, config):
+        if config['token']:
+            session.headers.update({'X-Auth-Token': config['token']})
+        if config['cert']:
+            session.cert = config['cert']
+
+    @classmethod
     def _create_session(cls, config):
         # allow endpoint path to not end with /
         # (handle incorrect user input when merging paths, see rfc3986, sec 5.2.3)
@@ -356,10 +363,7 @@ class DWaveAPIClient:
             session.headers.update(config['headers'])
 
         # auth
-        if config['token']:
-            session.headers.update({'X-Auth-Token': config['token']})
-        if config['cert']:
-            session.cert = config['cert']
+        cls._add_auth(session, config)
 
         if config['proxies']:
             session.proxies = config['proxies']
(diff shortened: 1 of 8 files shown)


dwave/cloud/api/client.py
```diff
@@ -345,3 +349,16 @@ class MetadataAPIClient(DWaveAPIClient):
...
         super().__init__(**config)
+
+
+class LeapAPIClient(DWaveAPIClient):
+    """Client for D-Wave's Leap API."""
+
+    @classmethod
+    def _add_auth(cls, session, config):
+        if config['token']:
+            session.headers.update({'Authorization': f"Bearer {config['token']}"})
```
arcondello, line 360: Bike shedding, I generally find `.update(Authorization=...)` a bit easier to read, and it also avoids the creation of an intermediate dictionary.

arcondello (approved):
LGTM

dwave/cloud/api/client.py
randomir: I generally prefer kwargs as well, but HTTP headers often contain hyphens (like `Accept-Language`), so I prefer string keys in that case.

dwave/cloud/api/resources.py
```diff
@@ -275,3 +276,39 @@ def cancel_problems(self, problem_ids: List[str]) -> \
...
         return parse_obj_as(rtype, response.json())
+
+
+class LeapAccount(ResourceBase):
+
+    resource_path = 'account/'
+    client_class = LeapAPIClient
+
+    # TODO: constrain accepted resource/response version, when
+    # media type defined for Leap API responses
```
randomir, line 287: @charleswhchan, I'll create an internal ticket for this.

Version decorators look, for example, like this:

https://github.com/dwavesystems/dwave-cloud-client/blob/d1bf2ee65e05bb974a7a2ef402c5a2413d060ae5/dwave/cloud/api/resources.py#L105-L106

dwave/cloud/api/client.py
```diff
@@ -326,6 +326,13 @@ def _retry_config(backoff_max=None, **kwargs):
 
         return retry
 
+    @classmethod
```
arcondello, line 329: Looks like this could be a `staticmethod`, though perhaps it makes sense to keep it consistent with some of the other `classmethod`s around.

arcondello (approved):
LGTM

dwave/cloud/api/client.py
randomir: I still haven't decided if this should actually be a regular method, with access to `self.config`, or not. That's why I kept it similar to the helper methods around. And private, so it's easy to change later.

charleswhchan:
I believe Bulat might have some input for this feature. Can you add him to the PR as well?

charleswhchan:
We shouldn't release this until zero downtime for Leap is completed (Q3 '23). Otherwise, calls to /leap will end up failing during by weekly deployment. Bad UX.

randomir:
> We shouldn't release this until zero downtime for Leap is completed (Q3 '23). Otherwise, calls to /leap will end up failing during by weekly deployment. Bad UX.

Aren't SAPI calls failing during SAPI deploys?!

charleswhchan:
sapi-ws has zero downtime implemented and scheduler has a very short delay before restart, so all /sapi calls should be handled and processed as usual.

where as calls to /leap will show the maintenance page and probably return a 5xx error

randomir:
We already use Leap API from the IDE, so I think it's fine to release this before end of Q3. It's not like it will be visible and used. Currently it's not even documented, so intended more for internal use.. for now.

arcondello:
> We shouldn't release this until zero downtime for Leap is completed (Q3 '23). Otherwise, calls to /leap will end up failing during by weekly deployment. Bad UX.

My 2c is that a user who is using this feature (especially given its current semi-private nature) would understand that when Leap is down they would get an error message. So IMO while it's not an ideal user experience, the relative benefits are well worth it.

charleswhchan:
> My 2c is that a user who is using this feature (especially given its current semi-private nature) would understand that when Leap is down they would get an error message. So IMO while it's not an ideal user experience, the relative benefits are well worth it.

Rad, Bulat and myself will be meeting on Monday -- we can reach a decision then.

charleswhchan (approved):

PR #541, opened 2023-02-22
Enforce API response version validation

randomir:
Implement D-Wave API response _data format_ and _version_ validation for every REST API resource method implemented in `dwave.cloud.api.resources.*`, i.e. `Regions`, `Solvers` and `Problems`.

For now, don't be strict -- skip the check if the response omits `Content-Type` (media type). We might want to implement a strict mode later.

In case the client receives an unsupported response version, we fail with `ResourceBadResponseError`, prompting the user to upgrade the client.

### DIFF
--- a/dwave/cloud/api/client.py
+++ b/dwave/cloud/api/client.py
@@ -14,9 +14,12 @@
 
 import logging
 from collections import deque, namedtuple
+from typing import Optional
 
 import requests
 import urllib3
+from packaging.specifiers import SpecifierSet
+from werkzeug.http import parse_options_header
 
 from dwave.cloud.api import constants, exceptions
 from dwave.cloud.package_info import __packagename__, __version__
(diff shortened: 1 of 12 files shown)


dwave/cloud/api/client.py
```diff
@@ -110,8 +113,89 @@ def request(self, method: str, *args, **kwargs):
...
+
+            version = params.pop('version', None)
+            if version is not None:
+                if not ss.contains(version):
+                    raise exceptions.ResourceBadResponseError(
+                        f'API response format version {version!r} not compliant '
+                        f'with supported version {self._accept_version!r}. '
+                        'Try upgrading "dwave-cloud-client".')
+
+    def request(self, method, url, *args, **kwargs):
```
arcondello, line 181: Not that it necessarily needs it, but inconsistent with the typing.

dwave/cloud/api/client.py
```diff
@@ -110,8 +113,89 @@ def request(self, method: str, *args, **kwargs):
...
+        headers = kwargs.pop('headers', {})
```
arcondello, line 182: Should this make a copy? It looks like `headers` might be modified below in 192.

arcondello (approved):
LGTM

dwave/cloud/api/client.py
randomir: Yes, that's a good catch, thanks.

Issue #426, opened 2020-09-17
Use a hierarchical config format


randomir:
We could switch from Python's config INI format to YAML (or JSON), enabling easy nesting of options.

For example:
```
client: qpu
solver:
  num_active_qubits:
    gt: 4000
  topology:
    type: pegasus
```

This would enable a partial update (`PATCH` with a diff), e.g:
```
solver:
  topology:
    type: chimera
```

PR #546, opened 2023-05-16
Update to use newer QPUs

JoelPasvolsky:
I mostly stuck to updating the examples with D-Wave 2000Q to Advantage and resisted restructuring and rewriting the surroundings wholesale, except a few places where I was overcome.

### DIFF
--- a/dwave/cloud/config.py
+++ b/dwave/cloud/config.py
@@ -65,9 +65,11 @@ Examples:
 
     *   **Example** Typical use for QPU and hybrid solvers
 
-    This example uses the following standard configuration file, 
-    :code:`~/.config/dwave/dwave.conf`, selected through auto-detection as the 
-    default configuration::
+    This example uses a configuration file in one of the standard paths, 
+    :code:`~/.config/dwave/dwave.conf`, selected through auto-detection
+    on the local system following the user/system configuration paths of 
+    :func:`get_configfile_paths`, to provide the following default 
+    configuration::
 
         [defaults]
         token = ABC-123456789123456789123456789
@@ -81,23 +83,22 @@ Examples:
         [europe]
         region = eu-central-1
     
-    That configuration files sets a default API token used by all following 
-    profiles unless overridden, ensures that by default selecting a solver with 
-    the :func:`~dwave.cloud.client.Client.get_solver` method returns a 
+    This configuration file sets a default API token used by all following 
+    profiles unless overridden, ensures that selecting a solver with the 
+    :func:`~dwave.cloud.client.Client.get_solver` method by default returns a 
     QPU solver with at least 5000 qubits, and configures profiles for selecting 
-    a quantum-classical hybrid solver and solvers from the European region.
+    a quantum-classical hybrid solver and solvers from Leap's European region.
 
-    The code below instantiates clients for a D-Wave quantum computer and a 
-    quantum-classical hybrid CQM solver.
+    The code below instantiates clients for a QPU solver and a hybrid CQM solver.
 
     >>> with Client.from_config() as client:   # doctest: +SKIP
     ...     solver_qpu = client.get_solver()
     >>> with Client.from_config(profile='hybrid') as client:   # doctest: +SKIP
     ...     solver_cqm = client.get_solver(supported_problem_types__issubset={"cqm"})
 
-    *   **Example:** Explicitly specified configuration file
+    *   **Example:** Explicitly specified configuration file, unrecognized parameter
     
-    This example explicitly specifies a configuration file, 
+    This example explicitly specifies the following configuration file, 
     :code:`~/jane/my_path_to_config/my_cloud_conf.conf`::
 
         [defaults]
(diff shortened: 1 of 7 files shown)


dwave/cloud/config.py
```diff
@@ -48,171 +48,141 @@
...
+        [defaults]
+        token = ABC-123456789123456789123456789
 
-    ``DWAVE_API_TOKEN``: API authorization token.
+        [default-solver]
+        solver = {"qpu": true, "num_qubits__gt": 5000}
 
-    ``DWAVE_API_SOLVER``: Default solver.
+        [hybrid]
+        client = hybrid
```
randomir, line 81: Probably more useful config would be:
```suggestion
        [bqm]
        client = hybrid
        solver = {"supported_problem_types__contains": "bqm"}

        [cqm]
        client = hybrid
        solver = {"supported_problem_types__contains": "cqm"}
```
(`client` is optional above, but can leave it for explicitness)

Then:
```
>>> with Client.from_config(profile="cqm") as client:
...     cqm_solver = client.get_solver()
```

dwave/cloud/config.py
```diff
@@ -48,171 +48,141 @@
...
     >>> import os
-    >>> os.environ['DWAVE_API_SOLVER'] = 'EXAMPLE_2000Q_SYSTEM'   # doctest: +SKIP
+    >>> os.environ["DWAVE_API_SOLVER"] = "Advantage_Emulator"   # doctest: +SKIP
     >>> dwave.cloud.config.load_config("./dwave_c.conf")   # doctest: +SKIP
-    {'client': 'sw',
-     'endpoint': 'https://url.of.some.dwavesystem.com/sapi',
-     'proxy': None,
-     'headers': None,
-     'solver': 'EXAMPLE_2000Q_SYSTEM',
+    {'solver': '{"software": true, "name": "Advantage_Emulator"}',
```
randomir, line 182: Haven't checked, but this should be:
```suggestion
    {'solver': 'Advantage_Emulator'},
```
If it's not, that might be a bug. Solver from the env should replace the complete solver definition, not just patch partially (name).

dwave/cloud/config.py
```diff
@@ -48,171 +48,141 @@
...
+    https://eu-central-1.cloud.dwavesys.com/sapi/v2/
+    >>> print(client.token)  # doctest: +SKIP
+    DEF-987654321987654321987654321
     >>> # code that uses client
     >>> client.close() # doctest: +SKIP
 
-    The next example uses :func:`~dwave.cloud.config.load_config` to load profile values.
-    **Most users do not need to use this method.** It loads from the following configuration
-    file, dwave_c.conf, located in the current working directory, and specified explicitly::
+    *   **Advanced Example:** Loading profile values
```
randomir, line 153: I would just drop this complete section. Looks too complicated to be useful, as it's artificial and not describing a use case someone might have in practice (you would never override the envvar from code when you can just use kwargs to client init).

dwave/cloud/config.py
```diff
@@ -48,171 +48,141 @@
...
-     'solver': 'EXAMPLE_2000Q_SYSTEM',
+    {'solver': '{"software": true, "name": "Advantage_Emulator"}',
      'token': 'ABC-123456789123456789123456789'}
-    >>> dc.config.load_config("./dwave_c.conf", profile='dw2000b', solver='Solver3')   # doctest: +SKIP
-    {'client': 'qpu',
-     'endpoint': 'https://url.of.some.dwavesystem.com/sapi',
-     'proxy': None,
-     'headers': None,
-     'solver': 'Solver3',
+    >>> dwave.cloud.config.load_config("./dwave_c.conf", profile="advantage-b", solver="Advantage_system4.1")   # doctest: +SKIP
```
randomir, line 184: I don't want to encourage any user to try something like this (use `load_config`). At least, not with the current client/config structure.

dwave/cloud/config.py
```diff
@@ -490,29 +460,27 @@ def load_config_from_files(filenames=None):
     Examples:
         This example loads configurations from two files. One contains a default
         section with key/values that are overwritten by any profile section that
-        contains that key/value; for example, profile dw2000b in file dwave_b.conf
-        overwrites the default URL and client type, which profile dw2000a inherits
-        from the defaults section, while profile dw2000a overwrites the API token that
-        profile dw2000b inherits.
+        contains that key/value; for example, profile advantage_b in file dwave_b.conf
```
randomir, line 463: Can we please just drop this whole example? The `load_config_from_files` is such a low-level internal function we don't want anyone ever use it :smile: 

OTOH, we don't have to modify docs significantly here, as the whole config module will be replaced in #426.

dwave/cloud/config.py
JoelPasvolsky: Dropped

dwave/cloud/config.py
JoelPasvolsky: Dropped example

dwave/cloud/config.py
JoelPasvolsky: Eliminated

PR #548, opened 2023-05-23
Fix partial resource decoration with `@accepts`

### LINKED ISSUE #544: Allow partial decoration of `Resource` methods with `@accepts`
Currently, all methods of a particular Resource class have to be decorated with `accepts`, or none. Mixing is not allowed.

randomir:
Fix #544.


arcondello (approved):
LGTM

PR #556, opened 2023-07-25
Upgrade models for pydantic v2

### LINKED ISSUE #555: Upgrade to pydantic 2+
Version 2 is out, and there are some major changes we need to adapt for.

Also see #553.

randomir:
Close #555.


randomir:
Python 3.10.0 failures caused by [a bug in pydantic](https://github.com/pydantic/pydantic/issues/6912).

hbarovertwo:
Just ran into the pydantic issue on python 3.8

dwave/cloud/utils.py
```diff
@@ -241,6 +242,24 @@ def strtrunc(s, maxlen=60):
     return s[:(maxlen-3)]+'...' if len(s) > maxlen else s
 
 
+def coerce_numpy_to_python(obj):
+    """Numpy object serializer with support for basic scalar types and ndarrays."""
+
+    if isinstance(obj, numpy.integer):
```
arcondello, line 248: I have a vague recollection that `isinstance` doesn't always work as expected for NumPy dtypes. There is [np.issubdtype](https://numpy.org/doc/stable/reference/generated/numpy.issubdtype.html) that is more forgiving. Unfortunately I don't remember the specific edge case.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 requests[socks]>=2.18
-pydantic>=1.7.3,<2
+pydantic>=2,<3
```
arcondello, line 2: I think you want to add `typing_extensions` to the `requirements.txt` as well.

arcondello (approved):
LGTM

dwave/cloud/utils.py
randomir: `np.issubdtype` would not be a good fit here, as it would pass types, not just objects (type instances). The next line `int(obj)` would fail in that case. We have pretty comprehensive [tests](https://github.com/dwavesystems/dwave-cloud-client/blob/2d537255fefc80dc2d061d9381019abdb8f4a377/tests/test_utils.py#L181-L231) for this function (they cover all NumPy scalar types, and a few array types of interest), and `ininstance` seems to work as expected.

requirements.txt
randomir: Good catch. Although it's already required by pydantic and the package, and for py38 only. :smiley:

PR #558, opened 2023-08-01
Drop Python 3.7 support and refresh CI

randomir:
In this PR we:
- drop py37 support
- update CI: always run the latest patch version of python (for each minor). Even on Windows! :smile: 
- drop appveyor; run windows tests on circleci


.circleci/config.yml
```diff
@@ -128,6 +127,46 @@ jobs:
...
+
+    steps:
+      - checkout
+
+      - run:
+          name: Install python and create virtualenv
+          shell: bash -eo pipefail
+          command: |
+            # resolve python MAJOR.MINOR version to latest MAJOR.MINOR.PATCH version available via NuGet
+            full_version=$(curl -s 'https://azuresearch-usnc.nuget.org/query?q=python' | jq '.data[] | select(.id=="python") .versions[] | .version' -r | awk -F. -v ver='<< parameters.python-version >>' 'index($0,ver".") == 1 && $3 > m { m=$3 } END { print ver "." m }')
```
randomir, line 146: This is why we should stop supporting windows! :laughing:

PR #568, opened 2023-09-01
Consolidate config parsing and validation

### LINKED ISSUE #504: Completely decouple config parsing from `Client` construction
Recently refactored `dwave.cloud.config.load_config()` is in a good shape, but we want to completely decouple config parsing from `Client` construction, so that config can be loaded independently and used outside of `Client` -- for example in `dwave.cloud.api.client.DWaveAPIClient`, its subclasses and by `dwave.cloud.api.resources.*`.

This should be combined with #426. And then #490, #440, #357 and other config improvements can follow.

### LINKED ISSUE #505: Use full client config for `dwave.cloud.api.Regions` requests
Currently we only use `metadata_api_endpoint` and `headers` because we need to resolve regions mid `Client` construction, i.e. config parsing.

Blocked by #504.

### LINKED ISSUE #507: Client init fails in 0.9.2 if proxy/permissive_ssl is required
We currently resolve region (using Metadata API) during `Client` construction. If `proxy` or `permissive_ssl` options are required for Metadata API requests (in addition to SAPI requests), region resolution (and Client init) will fail as those options are not parsed in time.

Implementing #505 will fully resolve this, but we might want to address it with a partial fix sooner.

### LINKED ISSUE #572: Add `from_config`-like factories to `dwave.cloud.api` clients and resources
After config loading is decoupled from `Client` init (#504), we can use it in the low-level API client and resource classes to simplify construction and usage.

randomir:
This PR primarily addresses #504 (decoupling config parsing/validation from `Client`), and then adds `from_config` factories to low-level API clients and resources (#572).

To simplify config parsing, validation and usage going forward we also define a new `dataclass`-like model for the configuration, `ClientConfig` (Pydantic-based).

We also prepare ground for config file v2, a hierarchical config format (#426).

Close #504.
Close #572.
Close #505.
Fix #507.


dwave/cloud/api/resources.py
```diff
@@ -86,16 +89,9 @@ def __exit__(self, exc_type, exc_value, traceback):
...
-        """Create Resource instance configured from a
-        :class:`~dwave.cloud.client.base.Client' instance.
-        """
-        # TODO: also accept configuration dict/dataclass when config/client refactored
-        if isinstance(client, DWaveAPIClient):
-            return cls(**client.config)
-        else: # assume isinstance(client, dwave.cloud.Client), without importing
-            sapiclient = SolverAPIClient.from_client_config(client)
-            return cls(**sapiclient.config)
+    def from_config(cls, config=None, **kwargs):
```
thisac, line 92: Missing type hints.

dwave/cloud/api/client.py
```diff
@@ -298,10 +302,79 @@ class DWaveAPIClient:
...
+        return cls.from_client_config(config)
+
+    @classmethod
+    def from_config(cls, config: Optional[Union[ClientConfig, str]] = None, **kwargs):
+        """Create class instance based on client config or config file.
+
+        Args:
+            config:
+                Path to configuration file or client config model. Based on `config`
+                type, we dispatch object creation to either :meth:`.from_config_file`
```
JoelPasvolsky, line 364: "we dispatch" --> "dispatches"

I'm mostly not looking at the docstrings but otherwise I'm always cleaning these up

dwave/cloud/config/models.py
```diff
@@ -0,0 +1,303 @@
...
+
+    The inverse might not be exact due to field type casting during validation.
+    E.g. original boolean value of 'off' or '0' will be dumped as 'false'.
+    """
+
+    raw_config = config.model_dump()
+
+    # invert transformations during validation
+    if h := raw_config['headers']:
+        raw_config['headers'] = '\n'.join("{k}: {v}" for k, v in h.items())
```
thisac, line 233: I assume this should be the values of `k` and `v` (using an f-string)?
```suggestion
        raw_config['headers'] = '\n'.join(f"{k}: {v}" for k, v in h.items())
```

dwave/cloud/api/client.py
```diff
@@ -298,10 +302,79 @@ class DWaveAPIClient:
...
+                or :meth:`.from_client_config`. If omitted, config load from file is
+                attempted.
```
JoelPasvolsky, line 366: "config load from file" --> can we just add "(the default)" to "to either :meth:`.from_config_file`" in the previous sentence (otherwise "If omitted, attempts to load configuration from file")

Okay, now I'll try not to look at more docstrings

dwave/cloud/config/models.py
```diff
@@ -0,0 +1,303 @@
...
+
+    # retry
+    prefix = 'http_retry_'
+    config['request_retry'] = {k[len(prefix):]: v for k, v in config.items()
+                               if k.startswith(prefix)}
+
+    return ClientConfig.model_validate(config)
+
+
+def dump_config_v1(config: ClientConfig) -> dict:
```
thisac, line 218: This function doesn't seem to be tested anywhere. Should it be?

thisac (approved):
LGTM, as far as I can tell after a quick review. Just a few comments.

dwave/cloud/config/models.py
randomir: Great catch, thanks.

dwave/cloud/config/models.py
randomir: I added just a "smoke test" for now, as we're not using this function (for now), but yeah, I'll add the full test -- the f-string bug above was convincing :smile:

JoelPasvolsky (approved):
LGTM.
Noting that you promised "We can update docs in a follow-up PR". For posterity.

dwave/cloud/api/client.py
```diff
@@ -298,10 +302,79 @@ class DWaveAPIClient:
     def __init__(self, **config):
         self.config = {}
         for opt, default in self.DEFAULTS.items():
-            self.config[opt] = config.get(opt, default)
+            val = config.get(opt)
+            if val is None:
+                val = default
+            self.config[opt] = val
```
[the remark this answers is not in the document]
randomir: To handle cases when key is present in dict, but value is `None`. Previously in `config` we had only manually specified kwargs, and now we expect to get the full `ClientConfig`, as dict (where we set optional fields to `None`).

dwave/cloud/api/client.py
```diff
@@ -298,10 +302,79 @@ class DWaveAPIClient:
...
+
+        # context-sensitive config update
+        update_config(opts, kwargs)
+
+        return cls(**opts)
+
+    @classmethod
+    def from_config_file(cls, config_file: Optional[str] = None,
+                         profile: Optional[str] = None, **kwargs):
+        """Create class instance based on config file / environment / kwarg options
```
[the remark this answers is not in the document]
randomir: `load_config()` few lines below actually loads from file and env. What's confusing is this method's name, but this is the best name I could come up with. I don't expect this method to get much usage.. The one you'll want to use is `from_config` -- which dispatches config loading to this method. And then the name (`from_config`) and the behavior is consistent with `Client.from_config`.

dwave/cloud/config/models.py
thisac: Hehe, sounds good.

randomir: Added tests for `dump_config_v1` in 60b3a9e6d2d573ebd9348176e1c9b66e849a41e0.

PR #575, opened 2023-09-13
Add timing info to `dwave sample`

### LINKED ISSUE #570: Add timing output to `dwave sample`
It's useful, and also makes the command output consistent with `dwave ping`.

randomir:
Close #570.

### DIFF
--- a/dwave/cloud/cli.py
+++ b/dwave/cloud/cli.py
@@ -584,6 +584,7 @@ def sample(*, config_file, profile, endpoint, region, client_type, solver_def,
         endpoint=endpoint, region=region,
         client=client_type, solver=solver_def)
 
+    t0 = timer()
     client, solver = _get_client_solver(config, output)
 
     if random_problem:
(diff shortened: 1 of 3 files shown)


dwave/cloud/cli.py
```diff
@@ -584,6 +584,7 @@ def sample(*, config_file, profile, endpoint, region, client_type, solver_def,
         endpoint=endpoint, region=region,
         client=client_type, solver=solver_def)
 
+    t0 = timer()
```
JoelPasvolsky, line 587: Consider making a timer dict so that instead of t0, t1 etc today and then possibly t0.5 tomorrow, you can have `{'start': ?, 'fetch': ? ...}`

releasenotes/notes/add-timing-output-to-dwave-sample-e3adb8b197921e8d.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Add wall-clock and QPU timing details to ``dwave sample`` output.
```
JoelPasvolsky, line 4: If your hyphenated "wall-clock" is an adjective on details then so should be "QPU-timing". Perhaps "Add wall time and QPU timing to ``dwave sample`` output."

JoelPasvolsky (approved):

PR #576, opened 2023-09-13
Support (read,connect) timeout tuple for `request_timeout` config param

### LINKED ISSUE #440: Split request_timeout config option to read and connect timeout
Needed for #439, but also to allow for finer granularity of timeouts.

We could introduce two new options and/or support a tuple for `request_timeout` (similar to the requests lib).

See also #357.

randomir:
Close #440.

### DIFF
--- a/dwave/cloud/config/models.py
+++ b/dwave/cloud/config/models.py
@@ -12,14 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import ast
 import copy
 import json
 import logging
 from collections import abc
 from typing import Optional, Union, Literal, Tuple, Dict, Any
+from typing_extensions import Annotated     # backport for py37, py38
 
 import urllib3
-from pydantic import BaseModel
+from pydantic import BaseModel, BeforeValidator
 
 from dwave.cloud.config.loaders import update_config
 from dwave.cloud.api.constants import (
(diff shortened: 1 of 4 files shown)


dwave/cloud/config/models.py
```diff
@@ -99,6 +101,12 @@ class PollingSchedule(BaseModel, GetterMixin):
     backoff_base: Optional[float] = 1.3
 
 
+def _literal_eval(obj):
+    if isinstance(obj, str):
+        return ast.literal_eval(obj)
```
arcondello, line 106: Didn't know about this, neat

arcondello (approved):
LGTM

dwave/cloud/config/models.py
```diff
@@ -99,6 +101,12 @@ class PollingSchedule(BaseModel, GetterMixin):
...
+    return obj
+
```
JoelPasvolsky, line 108: Is this a leftover? I don't see it used (plus line 15)

JoelPasvolsky (approved):

dwave/cloud/config/models.py
randomir: it's used in line 140 to define field validator

JoelPasvolsky: Ah, there you go, thanks!

dwave/cloud/config/models.py
randomir: Yup, we use it to safely parse linear/quad list/dict in [`dwave sample`](https://github.com/dwavesystems/dwave-cloud-client/blob/68078b1ca19f5534d36b469ac968eaf1c6e6322f/dwave/cloud/cli.py#L606) as well.

PR #580, opened 2023-09-22
Minor clean-ups of config and cache

randomir:


dwave/cloud/api/client.py
```diff
@@ -312,11 +312,11 @@ def __init__(self, **config):
         self.session = self._create_session(self.config)
 
     @classmethod
-    def from_client_config(cls, config: ClientConfig, **kwargs):
-        """Create class instance based on
+    def from_config_model(cls, config: ClientConfig, **kwargs):
```
arcondello, line 315: I wonder if these `from_config_X` methods could all be replaced by `.from_config()` and `functools.singledispatchmethod`. I know we do a very similar pattern in other parts of Ocean but I can't say that I've ever found it to be aesthetically satisfying.

There may be other considerations I am missing of course.

dwave/cloud/utils.py
```diff
@@ -546,15 +546,15 @@ def _argshash(self, args: List[Any], kwargs: Dict[Any, Any]):
 
     def __init__(self, *,
                  maxage: Optional[float] = None,
-                 cache: Optional[Mapping] = None,
```
arcondello, line 549: Do we want to still support `cache` with a deprecation warning?

Probably too low level for it to matter I suppose.

arcondello (approved):
LGTM!

dwave/cloud/utils.py
randomir: I expect this to have ~zero impact on users. It's not just semi-private (undocumented), but also you don't need to set `cache` param except for testing or exotic use cases...

dwave/cloud/api/client.py
randomir: I tried exactly that when splitting config into these submethods (which should see very rare use) in #568, but couldn't get the no-argument option (default `config=None`) working with `singledispatchmethod`.

PR #582, opened 2023-09-24
Add http server to handle auth callback

### LINKED ISSUE #563: Implement a utility HTTP server/interface
Local web server is needed to handle redirects from Leap and exchange the authorization code for token (see #561), without further user intervention (like manually pasting the code).

We already implement a simple web server in https://github.com/dwavesystems/dwave-inspector, one that is based on the built-in [`WSGIServer`](https://docs.python.org/3/library/wsgiref.html#wsgiref.simple_server.WSGIServer) and [`WSGIRequestHandler`](https://docs.python.org/3/library/wsgiref.html#wsgiref.simple_server.WSGIRequestHandler), and that it runs Flask.

We might want to either:
- move the dwave-inspector server implementation over to the cloud-client (since it's a dependency anyway); or
- implement a new server based on off-the-shelf `werkzeug.run_simple`, or similar.

randomir:
Close #563.

### DIFF
--- a/dwave/cloud/auth/flows.py
+++ b/dwave/cloud/auth/flows.py
@@ -106,6 +106,26 @@ class AuthFlow:
         return url
 
     def fetch_token(self, *, code: str, **kwargs) -> dict:
+        """Exchange ``code`` for access token.
+
+        Args:
+            code:
+                OAuth 2.0 authorization code to be exchanged for access token.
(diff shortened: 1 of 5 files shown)


arcondello (approved):
LGTM, though this is way out of my wheelhouse unfortunately.

PR #584, opened 2023-10-01
Improve auth flow

randomir:
In this PR we:
- make Ocean client id external configurable (via new config option `leap_client_id`)
- fix `LeapAuthFlow` requirements on endpoints available in config (they are now auto-resolved)
- fix caching of regions metadata to be persistent over program runs, also increase default max age to 7 days
- expose/proxy token handling methods on `AuthFlow`

### DIFF
--- a/dwave/cloud/client/base.py
+++ b/dwave/cloud/client/base.py
@@ -224,6 +224,10 @@ class Client(object):
         leap_api_endpoint (str, optional):
             Leap API endpoint.
 
+        leap_client_id (str, optional):
+            Leap OAuth 2.0 Ocean client id. Reserved for testing, otherwise
+            don't override.
+
         defaults (dict, optional):
             Defaults for the client instance that override the class
             :attr:`.Client.DEFAULTS`.
@@ -231,6 +235,9 @@ class Client(object):
     .. versionchanged:: 0.11.0
         Added the ``leap_api_endpoint`` parameter and config option (also
         available via environment variable ``DWAVE_LEAP_API_ENDPOINT``).
+
+        Added the ``leap_client_id`` parameter and config option (reserved for
+        testing).
 
     Note:
         Default values of all constructor arguments listed above are kept in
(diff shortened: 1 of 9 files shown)


dwave/cloud/client/base.py
```diff
@@ -224,6 +224,10 @@ class Client(object):
         leap_api_endpoint (str, optional):
             Leap API endpoint.
 
+        leap_client_id (str, optional):
+            Leap OAuth 2.0 Ocean client id. Reserved for testing, otherwise
```
arcondello, line 228: Maybe should be named `_leap_client_id` and not documented for now?

dwave/cloud/regions.py
```diff
@@ -22,18 +22,19 @@
...
+    DEFAULT_METADATA_API_ENDPOINT)
 from dwave.cloud.config.models import ClientConfig, validate_config_v1
 from dwave.cloud.utils import cached
 
 __all__ = ['get_regions']
 
 logger = logging.getLogger(__name__)
 
-_REGIONS_CACHE_MAXAGE = 86400   # 1 day
+_REGIONS_CACHE_MAXAGE = 7 * 86400   # 7 days
```
arcondello, line 34: Maybe one of these locations should import the same value from the other - single source of truth etc etc.

arcondello (approved):
LGTM

dwave/cloud/regions.py
randomir: They are independent. One is for general `get_regions` cache, and the other is specifically for calls from `Client`. The client can override the maxage, but for now I just kept them the same.

dwave/cloud/client/base.py
randomir: This is a config file option users will not care about, but it'll still be there. Unfortunately, `Client` class docs are currently the only place we document config options, and we should change that, especially with all the recent config changes.

PR #585, opened 2023-10-02
Add credentials storage

### LINKED ISSUE #562: Add credentials/tokens storage

randomir:
Close #562.

### DIFF
--- a/dwave/cloud/auth/creds.py
+++ b/dwave/cloud/auth/creds.py
@@ -13,15 +13,20 @@
 # limitations under the License.
 
 import logging
-from typing import List
+from pathlib import Path
+from typing import List, Optional, Union
+
+import diskcache
 
 from dwave.cloud.config import get_configfile_paths, get_default_configfile_path
 
-__all__ = []
+__all__ = ['Credentials']
 
 logger = logging.getLogger(__name__)
 
-CREDS_FILENAME = "credentials.db"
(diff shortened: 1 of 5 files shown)


dwave/cloud/auth/creds.py
```diff
@@ -0,0 +1,112 @@
...
+
+import diskcache
+
+from dwave.cloud.config import get_configfile_paths, get_default_configfile_path
+
+__all__ = ['Credentials']
+
+logger = logging.getLogger(__name__)
+
+CREDS_FILENAME = diskcache.core.DBNAME = "credentials.db"
```
arcondello, line 27: Wouldn't this interfere with another package also using `diskcache` by globally overloading the database name?

arcondello (approved):

dwave/cloud/auth/creds.py
randomir: It would. Not just another package, but it will affect our own `@cached.ondisk`. I will isolate this better. Unfortunately, there's no easy/nice way to isolate it..

randomir: Pushed an inelegant update (subclassed `diskcache.Cache`), but the only one I could see working robustly.

randomir: I'll submit a PR upstream with this change, and then we can remove it from Ocean.

PR #586, opened 2023-10-02
Add auth CLI

### LINKED ISSUE #565: Implement CLI support for auth (`dwave auth` subcommand)
Add `auth` subcommand to `dwave` CLI.

Add the following subcommands to `dwave auth`:
- `login` - initiate the OAuth flow. Use localhost-bound redirect URL, but support `--oob` (manual `code` & `state` entry)
- ~`list` - list existing identities/tokens~
- `get` - print access/refresh/id token
- `refresh` - refresh access/id token
- ~`revoke` - revoke a token~ (API not available)
- ~`ensure-valid` - refresh token only if needed~

randomir:
Close #565.

### DIFF
--- a/dwave/cloud/cli.py
+++ b/dwave/cloud/cli.py
@@ -47,7 +47,7 @@ from dwave.cloud.config import (
     get_configfile_path, get_default_configfile_path, get_configfile_paths)
 from dwave.cloud.config.models import validate_config_v1
 from dwave.cloud.regions import get_regions
-from dwave.cloud.auth.flows import LeapAuthFlow
+from dwave.cloud.auth.flows import LeapAuthFlow, OAuthError
 
 
 # show defaults for all click options when printing --help
@@ -324,11 +324,16 @@ def _config_create(config_file, profile, ask_full=False, auto_token=False):
         config.add_section(profile)
 
     # set using Leap API
-    sapi_token = _fetch_sapi_token(config_file, profile) if auto_token else None
+    try:
+        sapi_token = _fetch_sapi_token(config_file, profile) if auto_token else None
+    except Exception as error:
+        click.echo(f"Failed to fetch SAPI token from Leap API ({error!s}).")
+        sapi_token = None
+
     if sapi_token:
         del prompts['token']
         config.set(profile, 'token', sapi_token)
-        click.echo("Using SAPI token from credentials.")
+        click.echo("Using SAPI token fetched from Leap API.")
 
     _input_config_variables(config, profile, prompts)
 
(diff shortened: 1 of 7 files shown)


dwave/cloud/cli.py
```diff
@@ -292,6 +323,18 @@ def _config_create(config_file, profile, ask_full=False):
     if profile != default_section and not config.has_section(profile):
         config.add_section(profile)
 
+    # set using Leap API
+    try:
+        sapi_token = _fetch_sapi_token(config_file, profile) if auto_token else None
+    except Exception as error:
+        click.echo(f"Failed to fetch SAPI token from Leap API ({error!s}).")
```
JoelPasvolsky, line 330: Probably not this line but if I run `dwave auth login --profile test` and click `Decline` I get the error message `The provided authorization grant or refresh token is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client`, which is incorrect. Corner case that can be an issue for later, not in the least bit urgent.

dwave/cloud/cli.py
```diff
@@ -896,3 +943,156 @@ def setup(install_all, ask_full, verbose):
...
+    if not project:
+        raise CLIError(f'Project with {project_slug!r} ID, name or code not found. '
+                       'Please run "dwave leap project ls" to list available projects.', code=101)
+
+    # get token
+    project = project[0]
+    token = account.get_project_token(project=project)
+
+    output(f"Solver API token for project {project.name} ({project.code}) is {token}.",
+           token=token)
```
JoelPasvolsky, line 1098: I get an error for one particular project which is (1) last in the list (2) contains another project name but different case:

```
dwave leap project ls --profile test
Leap projects:
 - ... 
 - jp_test2 ...
 - tonytest  ...
 - TonyTestbeta ...
```

`dwave leap project token --project "tonytest" --profile=test` works but

```
dwave leap project token --project "TonyTestbeta" --profile=test
Unhandled error: 1 validation error for _LeapProjectTokenResponse
data.token
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.4/v/string_type
```

randomir: This is a validation error of the Leap API response -- the token for that project is `None`, which according to specs is incorrect. Now, if `None` is in fact allowed value for token returned from Leap API, I'll update expected response specs.

/cc @charleswhchan

dwave/cloud/cli.py
randomir: Good catch.

dwave/cloud/cli.py
randomir: Charles says this is a corner case for very old projects... I'll extend the model to support `null` tokens.

arcondello (approved):
LGTM

dwave/cloud/cli.py
randomir: @JoelPasvolsky, this should now be fixed, feel free to test it again.

PR #587, opened 2023-10-04
Tweak auth CLI

randomir:


dwave/cloud/cli.py
```diff
@@ -314,26 +297,34 @@ def _config_create(config_file, profile, ask_full=False, auto_token=False):
             click.echo(f"Available profiles: {', '.join(existing)}")
         default_profile = next(iter(existing))
 
-        _note = " (select existing or create new)" if config_file_exists else ""
-        profile = default_text_input(f"Profile{_note}", default_profile)
+        if ask_full:
+            _note = " (select existing or create new)" if config_file_exists else ""
```
arcondello, line 301: Maybe I can't see it, but does this get used anywhere? If no then it can probably be dropped. If yes, I think you need to specify a `_note` in the `else` block

dwave/cloud/cli.py
```diff
@@ -1065,38 +1053,57 @@ def list_leap_projects(*, config_file, profile, json_output, output):
...
-    needle = str(project_slug).strip().lower()
-    project = [p for p in projects if needle in (str(p.id), p.name.lower(), p.code.lower())]
+    if project_hint is None:
+        # use active project
+        project = account.get_active_project()
+
+    else:
+        # find project using project_hint
+        projects = account.list_projects()
+        needle = str(project_hint).strip().lower()
```
arcondello, line 1096: :laughing:  good variable name

arcondello (approved):

dwave/cloud/cli.py
randomir: You mean `_note` variable? It's used in the line below. And in case config file doesn't exist, note should be empty.

arcondello: So it is! I just can't read. Ignore me

dwave/cloud/cli.py
randomir: perfect for a 2-line scope :smile:

PR #593, opened 2023-10-27
Update auth flow pages

randomir:
Redirect to Leap success page (or Leap error page) upon redirect auth flow completion.

### DIFF
--- a/dwave/cloud/auth/flows.py
+++ b/dwave/cloud/auth/flows.py
@@ -19,12 +19,13 @@ import time
 import webbrowser
 from operator import sub
 from typing import Any, Callable, Dict, Optional, Union, Sequence
-from urllib.parse import urljoin, urlencode
+from urllib.parse import urljoin
 
 import click
 from authlib.integrations.requests_client import OAuth2Session, OAuthError
 from authlib.oauth2.rfc6749 import OAuth2Token
 from authlib.common.security import generate_token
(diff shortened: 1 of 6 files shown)


dwave/cloud/auth/flows.py
```diff
@@ -252,6 +253,14 @@ def _infer_auth_endpoint(leap_api_endpoint: str) -> str:
     def _infer_token_endpoint(leap_api_endpoint: str) -> str:
         return urljoin(leap_api_endpoint, '/leap/openid/token')
 
+    @staticmethod
+    def _infer_leap_success_uri(leap_api_endpoint: str) -> str:
+        return urljoin(leap_api_endpoint, '/leap/openid/success')
+
+    @staticmethod
+    def _infer_leap_error_uri(leap_api_endpoint: str) -> str:
+        return urljoin(leap_api_endpoint, '/leap/openid/error')
```
arcondello, line 262: I imagine that the answer is that these may eventually be used elsewhere, but why not just using `partial` in the `run_redirect_flow()` function body?

arcondello (approved):
LGTM

dwave/cloud/auth/flows.py
randomir: I hope to replace some of these by reading them from server metadata endpoint, once a couple of issues are fixed on SAPI.

PR #594, opened 2023-10-27
Gently prefer QPU solvers on `dwave ping`

### LINKED ISSUE #592: Prefer QPU solvers in `dwave ping`
As the CLI docs already claim (because they weren't updated after we added support for HSS/unstructured solvers).

There are two main issues with randomly picking a solver to ping:
- HSS solvers have required parameters (`time_limit`) client is not aware of, hence ping fails unless user specifies them with `-m/--sampling-params`. Once Solver API exposes more detailed solver parameter description, the client will be able to automatically pick a reasonable value for required params, and we can revisit this approach.
- HSS solvers have minimum runtime of at least 3s, hence sampling a dummy problem on them is slower than using a QPU solver

randomir:
Close #592.

### DIFF
--- a/dwave/cloud/cli.py
+++ b/dwave/cloud/cli.py
@@ -483,11 +483,18 @@ def ping(*, config_file, profile, endpoint, region, client_type, solver_def,
     if label:
         params.update(label=label)
 
+    # gently prefer the least busy QPU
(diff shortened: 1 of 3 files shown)


arcondello (approved):
Nice!

PR #596, opened 2023-10-30
Add Python 3.12 support

randomir:


dwave/cloud/utils.py
```diff
@@ -867,11 +873,52 @@ def get_contrib_packages():
...
     """Return a list of platform tags generated from registered entry points."""
 
-    fs = [ep.load() for ep in iter_entry_points('dwave.common.platform.tags')]
+    fs = [ep.load() for ep in entry_points(group='dwave.common.platform.tags')]
     tags = list(filter(None, [f() for f in fs]))
     return tags
 
 
+class VersionNotFoundError(Exception):
+    """Package version requirement not satisfied."""
```
JoelPasvolsky, line 882: Did you mean do expand this class definition?

dwave/cloud/utils.py
```diff
@@ -867,11 +873,52 @@ def get_contrib_packages():
...
+    Args:
+        requirement:
+            Package dependency requirement according to PEP-508, given as string,
+            or :class:`~packaging.requirements.Requirement`.
+        prereleases:
+            Boolean flag to control if installed prereleases are allowed.
+
+    Raises:
+        :class:`~importlib.metadata.PackageNotFoundError`:
+            Package by name not found.
```
JoelPasvolsky, line 903:
```suggestion
            Named package not found.
```

Is this the meaning?

dwave/cloud/utils.py
```diff
@@ -867,11 +873,52 @@ def get_contrib_packages():
...
+    fs = [ep.load() for ep in entry_points(group='dwave.common.platform.tags')]
     tags = list(filter(None, [f() for f in fs]))
     return tags
 
 
+class VersionNotFoundError(Exception):
+    """Package version requirement not satisfied."""
+
+
+def get_distribution(requirement: Union[str, Requirement],
```
JoelPasvolsky, line 885: Will [this functionality](https://github.com/dwavesystems/dwave-ocean-sdk/blob/fe16c75bf77583b78f6661754d1b1048b3704050/docs/conf.py#L213) be affected and should it be updated ([this PR](https://github.com/dwavesystems/dwave-ocean-sdk/pull/274) moves the doc build to py3.10)?

JoelPasvolsky (approved):

dwave/cloud/utils.py
randomir: No. I believe every package has a name. This error is raised when a search for package by name fails. The alternative to search-by-name is search by full requirement that includes name and version specifier.

dwave/cloud/utils.py
randomir: No, I think this is more than enough. Have in mind this is an internal, undocumented exception raised by an internal, undocumented function (`get_distribution`).

dwave/cloud/utils.py
randomir: That functionality is affected by the fact that Python 3.12 does not create virtual environments with `setuptools` preinstalled. `pkg_resources` is provided by `setuptools`. And that's the reason for some of these changes here - to support default py312 environments. One (lazy) alternative is to explicitly install `setuptools`. Another option is to switch to `importlib.resources` and `importlib.metadata`, depending on your requirements.

The way you use `pkg_resources.get_distribution` in Ocean's docs `conf.py` can easily be replaced with [`importlib.metadata.distribution`](https://docs.python.org/3/library/importlib.metadata.html#distributions).

PR #604, opened 2024-02-24
Fix RtD docs build

randomir:
Fix `reno` version traversal during RtD docs build (see error [here](https://readthedocs.com/projects/d-wave-systems-dwave-cloud-client/builds/2040024/)) by doing an [unshallow git clone](https://docs.readthedocs.io/en/stable/build-customization.html#unshallow-git-clone) (instead of the default shallow).

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -4,6 +4,9 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  jobs:
+    post_checkout:
+      - git fetch --unshallow || true
 
 python:
   install:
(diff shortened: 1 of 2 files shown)


.readthedocs.yaml
```diff
@@ -4,6 +4,9 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  jobs:
+    post_checkout:
+      - git fetch --unshallow || true
```
JoelPasvolsky, line 9: What needs the full git clone?

JoelPasvolsky (approved):

.readthedocs.yaml
randomir: `reno` does. It traverses git history to pull version numbers.

PR #608, opened 2024-02-29
Speed-up `User-Agent` construction

randomir:
Default UA we now use consistently across all clients (`base.Client`, `api.Client`, `auth.flow`) is now constructed in **under 5us instead of 11ms+** (on my machine).

Consequently, client construction is sped-up as well.

### DIFF
--- a/dwave/cloud/api/client.py
+++ b/dwave/cloud/api/client.py
@@ -26,30 +26,12 @@ from werkzeug.http import parse_options_header, dump_options_header
 from dwave.cloud.api import constants, exceptions
 from dwave.cloud.config import load_config, validate_config_v1, update_config
 from dwave.cloud.config.models import ClientConfig
-from dwave.cloud.package_info import __packagename__, __version__
 from dwave.cloud.utils import (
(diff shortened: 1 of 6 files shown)


dwave/cloud/api/client.py
```diff
@@ -293,11 +275,8 @@ class DWaveAPIClient:
     # client instance config, populated on init from kwargs overridding DEFAULTS
     config = None
 
-    # User-Agent string used in API requests, as returned by
-    # :meth:`~dwave.cloud.utils.user_agent`, computed on first access and
-    # cached for the lifespan of the class.
     # TODO: consider exposing "user_agent" config parameter
-    user_agent = LazyUserAgentClassProperty()
+    user_agent = default_user_agent()
```
arcondello, line 279: Clarifying question, you have a note later where you said you made `default_user_agent()` a function because it might change during runtime. But here you bind it on import to the class, so it won't change later. Does this get overriden somewhere else on instantiation or something like that?

randomir: good catch, I'll rather initialize it on (session) init.

arcondello (approved):

thisac (approved):

PR #609, opened 2024-03-05
Implement OAuth 2.0 token revocation

### LINKED ISSUE #600: Implement OAuth token revoke interface
- Add `revoke_token()` to `dwave.cloud.auth.AuthFlow`
- Add `dwave auth revoke` subcommand (as initially planned in #565).

randomir:
Close #600.

Note: token revocation is currently only implemented in the EXP environment.

### DIFF
--- a/dwave/cloud/auth/flows.py
+++ b/dwave/cloud/auth/flows.py
@@ -22,6 +22,7 @@ from typing import Any, Callable, Dict, Optional, Union, Sequence, Literal
 from urllib.parse import urljoin
 
 import click
+import requests
 from authlib.integrations.requests_client import OAuth2Session, OAuthError
 from authlib.oauth2.rfc6749 import OAuth2Token
 from authlib.common.security import generate_token
@@ -212,11 +213,29 @@ class AuthFlow:
         return token
 
     def revoke_token(self, *,
-                     token_type_hint: Optional[Literal['access_token',
-                                                       'refresh_token']] = None):
-        return self.session.revoke_token(
-            url=self.revocation_endpoint,
+                     token: Optional[str] = None,
+                     token_type_hint: Optional[
+                         Literal['access_token', 'refresh_token']] = None
+                     ) -> bool:
+        """Revoke OAuth 2.0 token.
+
+        Args:
+            token:
+                OAuth 2.0 access or refresh token to be revoked. If refresh token
+                is specified, the authorization server should also revoke the
+                corresponding access tokens. If ``token`` is not specified, then
+                current refresh token is revoked.
+            token_type_hint:
+                Type of the token to be revoked.
+
+        Returns:
+            Token revocation status. True if revocation was successful, or token
+            was invalid already. False in case of an error.
+        """
+        response: requests.Response = self.session.revoke_token(
+            url=self.revocation_endpoint, token=token,
             token_type_hint=token_type_hint)
+        return response.ok
 
     def ensure_active_token(self):
         if not self.token:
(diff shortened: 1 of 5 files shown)


dwave/cloud/cli.py
```diff
@@ -1051,6 +1051,32 @@ def refresh(*, config_file, profile, output):
...
+@standardized_output
+def revoke(*, config_file, profile, token_type, json_output, output):
+    """Revoke Leap API access and/or refresh token(s)."""
+
+    config = validate_config_v1(load_config(config_file=config_file, profile=profile))
+    flow = LeapAuthFlow.from_config_model(config)
+
+    token_key = token_type.replace('-', '_')
+    token_pretty = token_type.replace('-', ' ').capitalize()
+    if not flow.token or not (token_value := flow.token.get(token_key)):
```
arcondello, line 1068: Not to be Guido, but IMO if you're going to use `token_value` outside of the if block I would not use the walrus operator here. Took me a moment to find where it got assigned. I guess it might make sense if `flow.token.get(...)` is very expensive and you want to skip it if `not flow.token`, but IMO once you're throwing errors performance isn't as important.

arcondello (approved):
LGTM

dwave/cloud/cli.py
randomir: Agree, I'll fix it. The code looked differently when I wrote it like this.

dwave/cloud/auth/flows.py
```diff
@@ -205,6 +212,39 @@ def refresh_token(self):
...
+            token_type_hint:
+                Type of the token to be revoked.
+
+        Returns:
+            Token revocation status. True if revocation was successful, or token
+            was invalid already. False in case of an error.
+
+        Note:
+            Precondition for token revocation is defining a ``revocation_endpoint``
+            during :class:`~dwave.cloud.auth.flows.AuthFlow` construction.
```
JoelPasvolsky, line 237:
```suggestion
            Token revocation requires that a ``revocation_endpoint`` is defined
            in :class:`~dwave.cloud.auth.flows.AuthFlow` construction.
```

dwave/cloud/auth/flows.py
```diff
@@ -205,6 +212,39 @@ def refresh_token(self):
...
+                     token_type_hint: Optional[
+                         Literal['access_token', 'refresh_token']] = None
+                     ) -> bool:
+        """Revoke OAuth 2.0 token.
+
+        Args:
+            token:
+                OAuth 2.0 access or refresh token to be revoked. If refresh token
+                is specified, the authorization server should also revoke the
+                corresponding access tokens. If ``token`` is not specified, then
```
JoelPasvolsky, line 226: From "If ``token`` is not specified, then current refresh token is revoked" I'd understand that `revoke_token(token_type_hint='access_token')` will revoke the refresh token: is that correct?

dwave/cloud/cli.py
```diff
@@ -1051,6 +1051,32 @@ def refresh(*, config_file, profile, output):
...
 
 
+@auth.command()
+@config_file_options()
+@click.argument('token_type', default='access-token',
+                type=click.Choice(['access-token', 'refresh-token']))
+@json_output
+@standardized_output
+def revoke(*, config_file, profile, token_type, json_output, output):
+    """Revoke Leap API access and/or refresh token(s)."""
```
JoelPasvolsky, line 1061:
```suggestion
    """Revoke access and/or refresh token(s) for Leap API."""
```

Reading "Revoke Leap API access" will scare people

JoelPasvolsky (commented):
LGTM

JoelPasvolsky (approved):

thisac (approved):
LGTM

dwave/cloud/auth/flows.py
randomir: No, that's just a _hint_ for authorization server. If user doesn't specify `token`, the client library tries using refresh token, and if refresh token is not available (should never really happen here), then access token is used.

PR #625, opened 2024-06-11
Test logging configurations

randomir:
Here we update our logging config helper and test some interesting/useful configurations.

### DIFF
--- a/dwave/cloud/__init__.py
+++ b/dwave/cloud/__init__.py
@@ -98,6 +98,7 @@ def configure_logging(logger: typing.Optional[logging.Logger] = None,
                       output_stream: io.IOBase = sys.stderr,
                       in_utc: bool = False,
                       structured_output: bool = False,
+                      additive: bool = False,
                       ) -> logging.Logger:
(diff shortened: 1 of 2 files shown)


dwave/cloud/__init__.py
```diff
@@ -128,12 +137,18 @@ class Formatter(FilteredSecretsFormatter, formatter_base):
     else:
         Formatter = formatter_base
 
+    if not additive:
+        # make sure handlers are not accumulated
+        while len(logger.handlers):
```
arcondello, line 142: This makes me slightly nervous in case there are some edge cases where `removeHandler()` doesn't actually remove anything. But probably safe.

arcondello (approved):
LGTM

dwave/cloud/__init__.py
randomir: That function sounds more sophisticated than [it really is](https://github.com/python/cpython/blob/0315fdc24df9be03ae6ba40bac25b512e1e8da0a/Lib/logging/__init__.py#L1713). :slightly_smiling_face:

PR #627, opened 2024-06-12
Add nonlinear solver support

randomir:


dwave/cloud/solver.py
```diff
@@ -739,6 +749,149 @@ def upload_cqm(self, cqm):
...
+
+    Args:
+        client (:class:`~dwave.cloud.client.Client`):
+            Client that manages access to this solver.
+
+        data (`dict`):
+            Data from the server describing this solver.
+    """
+
+    _handled_problem_types = {"nl"}
```
arcondello, line 766: Almost certainly doesn't matter, but `frozenset` would be very slightly "safer"?

dwave/cloud/solver.py
```diff
@@ -739,6 +749,149 @@ def upload_cqm(self, cqm):
...
+                               "Re-install the library with 'nlm' support.")
+
+        # cqm to model generator requires binary bqm
+        bqm.change_vartype(dimod.BINARY, inplace=True)
+
+        # TODO: simplify when a public BQM -> Model generator is added to dwave-optimization
+        cqm = dimod.CQM.from_bqm(bqm)
+        nlm = _from_constrained_quadratic_model(cqm)
+
+        return self.sample_nlm(nlm, label=label, **params)
```
arcondello, line 808: Should this return a sample set? I guess if it's just for testing we could change it later.

arcondello (approved):
LGTM!

dwave/cloud/solver.py
randomir: No, all `Solver.sample_*` methods return `dwave.cloud.computation.Future`. I missed annotating the rtype here. Will fix.

dwave/cloud/solver.py
randomir: Probably. Or `tuple` even would probably be better.. But we should change it on all solvers then..

arcondello (approved):

Issue #437, opened 2020-10-30
General-use utils should have its own namespace


JoelPasvolsky:
Generally useful decorator like cloud-client's @dwave.cloud.utils.retried or @dwave_networkx.utils.decorators.binary_quadratic_model_sampler that aren't limited in use to the particular repo that were written in should be under a generic dwave.utils or dwave.decorators namespace.

arcondello:
I know that @randomir has been suggesting a `dwave.common` package or similar. Unfortunately, there is no nice way to populate a namespace from multiple packages (or at least none that I know of, though I haven't re-checked recently). https://github.com/dwavesystems/dwave-ocean-sdk/pull/21 is an attempt to create one using entry points.

randomir:
Yes, I still have `dwave.common` on my todo list. It would include:
- logging stuff
- profiling (like `tictoc`)
- testing utils, like various `assertX`
- misc commonly used utils like: `@retried`, `@cached`, `@deprecated`, etc.

Basically, utilities we use across multiple Ocean packages.

PR #628, opened 2024-06-13
Refactor/split utils into submodules

randomir:
Related to #437.

Enables import time optimization.


arcondello (approved):
I didn't look too closely at the functions because I believe most of them have just been moved. Structure seems like a nice improvement.

PR #632, opened 2024-06-17
Fix entry points enumeration across python versions

### LINKED ISSUE #631: `dwave install` broken on py39 after utils refactor
Due to a change in `entry_points()` behavior across a few versions of python (see compatibility note in [docs](https://docs.python.org/3/library/importlib.metadata.html#entry-points)), we use `entry_points()` from `importlib_metadata` backport, instead of `importlib.metadata`.

However, during utils refactor in #628, we switched to `importlib.metadata`. This causes failure of `get_contrib_config` (used from `dwave install` and `dwave setup` CLI commands) on Python 3.9 (only!).

https://github.com/dwavesystems/dwave-cloud-client/blob/30d69493a5f3b3fdaeeda872d253d5b735ad8fa0/dwave/cloud/utils/dist.py#L36

randomir:
Fix #631.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -213,6 +213,42 @@ jobs:
             . env/bin/activate
             make -C docs/ linkcheck
 
+  test-cli:
+    parameters:
+      python-version:
+        type: string
+
+    docker:
+      - image: cimg/python:<< parameters.python-version >>
+
+    steps:
+      - checkout
+
+      - run:
+          name: Install
+          command: |
+            python -m venv env
+            . env/bin/activate
+            pip install .
+            pip install dwave-inspector
+
+      - run:
+          name: Test contrib install
+          command: |
+            . env/bin/activate
+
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -298,6 +334,12 @@ workflows:
             parameters:
               python-version: *python-versions
 
+      - test-cli:
+          name: test-cli-linux-py<< matrix.python-version >>
+          matrix:
+            parameters:
+              python-version: &python-versions ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
thisac, line 341: Can't this use the same reference as above (`*python-versions`)?

thisac (approved):
LGTM

.circleci/config.yml
randomir: of course it can :smiley:  good catch, thanks!

PR #635, opened 2024-06-28
Add zephyr mocks

randomir:


dwave/cloud/testing/mocks.py
```diff
@@ -223,6 +223,44 @@ def qpu_pegasus_solver_data(m: int,
...
+            Solver properties passed down to :meth:`.structured_solver_data`.
+    """
+    try:
+        import dwave_networkx as dnx
+    except ImportError:     # pragma: no cover
+        raise RuntimeError("Can't generate Zephyr graph without dwave-networkx. "
+                           "Install with 'dwave-cloud-client[mocks]'.")
+
+    graph = dnx.zephyr_graph(m, t)
+    qubits = sorted(list(graph.nodes))
```
JoelPasvolsky, line 247: Why sort here? Looks like Zephyr nodes are ordered already

JoelPasvolsky (approved):
LGTM

dwave/cloud/testing/mocks.py
randomir: Yes, as are Pegasus nodes. But not Chimera.. So, just to ensure qubits and couplers are ordered like they are on SAPI (future-proof).

PR #640, opened 2024-07-22
Add airspeed velocity benchmarks

randomir:


benchmarks/fixtures/solvers.json
arcondello, line 1: I forget what asv is doing, what is this file?

arcondello (approved):

benchmarks/fixtures/solvers.json
randomir: All solvers metadata, as returned by SAPI. A fixture used to more realistically benchmark operations on solver data.

.circleci/config.yml
```diff
@@ -251,6 +251,22 @@ jobs:
...
+      - checkout
+
+      - run:
+          name: Verify benchmarks run
+          command: |
+            python -m venv env
+            . env/bin/activate
+            pip install -r benchmarks/requirements.txt
+            asv machine --yes
+            asv run -q
```
thisac, line 268: These are used here as plain functionality tests, right, and not for any actual performance degradation detection?

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,201 @@
...
+from dwave.cloud.config import ClientConfig
+from dwave.cloud.solver import StructuredSolver, BQMSolver, CQMSolver, DQMSolver, NLSolver
+from dwave.cloud.regions import resolve_endpoints, get_regions
+from dwave.cloud.testing import isolated_environ, mocks
+from dwave.cloud.utils.qubo import generate_random_ising_problem
+
+
+# note: looks like timeraw benchmarks can't be parameterized
+class ImportDeps:
+    version = "1"
```
thisac, line 32: Is there a need to add version to all the classes as well as in the conf file, or is this useful for some other reason?

thisac (approved):
Neat, have never had to use asv. Seems quite useful!

.circleci/config.yml
randomir: Exactly, just a functionality test in CI.

benchmarks/benchmarks.py
randomir: This is individual benchmark version (see `version` in [docs](https://asv.readthedocs.io/en/v0.6.3/benchmarks.html#benchmark-attributes)). I'm specifying it explicitly to be able to control it better. Otherwise a commit hash is used, and when a minor code change happens, previous results are dropped.

PR #643, opened 2024-07-27
Speed up imports

randomir:
Top-level namespace import is 20x faster; api and config imports around 30% faster; `dwave.cloud.Client` as before.

### DIFF
--- a/dwave/cloud/__init__.py
+++ b/dwave/cloud/__init__.py
@@ -43,6 +43,11 @@ def __getattr__(cls):
 
     raise AttributeError(f"module '{__name__}' has no attribute '{cls}'")
 
+def __dir__():
+    # dev note: make sure __all__ is limited to symbols we import lazily
+    yield from __all__
+    yield from globals()
+
 
 # lazily alias dwave.cloud.client.{qpu,sw,hybrid} as dwave.cloud.*
 # (i.e. don't preemptively import Client from submodules)
@@ -61,6 +66,10 @@ def _alias_old_client_submodules():
                 return getattr(mod, 'Client')
 
             raise AttributeError(f"module '{self.__name__}' has no attribute '{name}'")
+
+        def __dir__(self):
+            self.__getattr__('Client')      # force submodule import
+            yield from dir(sys.modules[f'dwave.cloud.client.{self.__name__}'])
 
     for name in ('qpu', 'sw', 'hybrid'):
         globals()[name] = sys.modules[f'dwave.cloud.{name}'] = _submod(name)
--- a/dwave/cloud/__init__.py
+++ b/dwave/cloud/__init__.py
@@ -15,10 +15,8 @@
 import sys
 import logging
 import importlib
+import types
 
-from dwave.cloud.client import Client
-from dwave.cloud.solver import Solver
-from dwave.cloud.computation import Future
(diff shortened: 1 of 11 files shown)


dwave/cloud/__init__.py
```diff
@@ -35,11 +34,36 @@
 configure_logging_from_env(logger)
 
 
-# alias dwave.cloud.client.{qpu,sw,hybrid} as dwave.cloud.*
+# lazily import Client/Solver/Future -- only when actually asked for
+def __getattr__(cls):
```
arcondello, line 38: Should we implement [`__dir__()`](https://docs.python.org/3/reference/datamodel.html#object.__dir__) as well? Might be anecdotal, but I use `dir(module)` fairly often when using a new library.

dwave/cloud/auth/flows.py
```diff
@@ -31,11 +31,13 @@
 from dwave.cloud.auth.config import OCEAN_SDK_CLIENT_ID, OCEAN_SDK_SCOPES
 from dwave.cloud.auth.creds import Credentials
 from dwave.cloud.auth.server import SingleRequestAppServer, RequestCaptureAndRedirectApp
-from dwave.cloud.config.models import ClientConfig
 from dwave.cloud.regions import resolve_endpoints
 from dwave.cloud.utils.http import default_user_agent
 from dwave.cloud.utils.logging import pretty_argvalues
 
+if TYPE_CHECKING:
```
arcondello, line 38: This is complaining about Python, not the code, but I find this such an awkward flow. I wish there was like a `timport` or something like that.

dwave/cloud/utils/decorators.py
```diff
@@ -330,6 +329,10 @@ def f(x):
         """
 
         def start(self):
+            # lazy import: cached.disabled is rarely used, yet mock import
+            # adds 50ms to root package import
```
arcondello, line 333: :exploding_head:

arcondello (approved):
LGTM!

dwave/cloud/__init__.py
randomir: I can add `__dir__()` but I think I would like to deprecate `from dwave.cloud import Client` form. This lazy getattr import is here to support the existing code, but I'm no sure I want to promote it further. :shrug:

arcondello: Fair enough. I do think at least `__dir__()` and `__all__` should be in sync, but flexible :shrug:

randomir: Fair point, I'll add it.

randomir: Added `__dir_()`. But this note in [docs](https://docs.python.org/3/library/functions.html#dir) is interesting:
> Note: Because [dir()](https://docs.python.org/3/library/functions.html#dir) is supplied primarily as a convenience for use > at an interactive prompt, it tries to supply an interesting set of names more than it tries to supply a rigorously or
> consistently defined set of names, and its detailed behavior may change across releases. For example, metaclass
> attributes are not in the result list when the argument is a class.

arcondello: For sure. I would never treat it as authoritative. I do think it's a useful convenience feature, in a prompt for instance, so nice to have.

arcondello (approved):

PR #645, opened 2024-07-31
Add solver metadata caching with validation via conditional requests

### LINKED ISSUE #613: Cache solver metadata
Once SAPI adds support for conditional requests ([rfc7232](https://datatracker.ietf.org/doc/html/rfc7232)), we can use it to avoid downloading of rather big solver metadata on each `Solver` init.

For this, we require:
- support for `ETag` and `If-None-Match`
- support on `/solvers/remote` and `/solvers/remote/<solver-id>` endpoints

randomir:
Close #613.

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -15,10 +15,15 @@
 import json
 import subprocess
 
+from copy import deepcopy
 from functools import partial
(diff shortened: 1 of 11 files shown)


arcondello (approved):
LGTM!

PR #649, opened 2024-08-01
Add long polling support

### LINKED ISSUE #648: Add long polling support
Recently SAPI added support for long polling calls on problem status endpoints.

Long polling behavior is activated by appending `timeout=<sec>` query parameter.

Default value for timeout is 1 second, and maximum allowed value is 30 seconds.

In addition to short polling with exponential back-off, we want to add long-polling as polling schedule strategy, and eventually make it the default.

randomir:
Close #648.

### DIFF
--- a/dwave/cloud/client/base.py
+++ b/dwave/cloud/client/base.py
@@ -307,10 +307,14 @@ class Client(object):
         'headers': None,
         'client_cert': None,
         'client_cert_key': None,
+        'poll_strategy': 'backoff',
         # poll back-off schedule defaults [sec]
         'poll_backoff_min': 0.05,
         'poll_backoff_max': 60,
         'poll_backoff_base': 1.3,
+        # long polling parameters
+        'poll_wait_time': 30,
+        'poll_pause': 0,
         # idempotent http requests retry params
         'http_retry_total': 10,
         'http_retry_connect': None,
--- a/dwave/cloud/client/base.py
+++ b/dwave/cloud/client/base.py
@@ -66,7 +66,7 @@ from dwave.cloud.package_info import __packagename__, __version__
 from dwave.cloud.exceptions import *    # TODO: fix
 from dwave.cloud.computation import Future
 from dwave.cloud.config import load_config, update_config, validate_config_v1
-from dwave.cloud.config.models import ClientConfig
+from dwave.cloud.config.models import ClientConfig, PollingStrategy
 from dwave.cloud.solver import available_solvers, StructuredSolver, UnstructuredSolver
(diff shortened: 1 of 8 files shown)


dwave/cloud/client/base.py
```diff
@@ -1449,9 +1453,16 @@ def _do_cancel_problems(self):
...
 
-    def _poll(self, future):
+    def _poll(self, future: Future) -> None:
         """Enqueue a problem to poll the server for status."""
 
+        # split for simplicity
+        if self.config.polling_schedule.strategy == PollingStrategy.BACKOFF:
+            return self._poll_using_backoff(future)
+        else:
+            return self._poll_using_long_polling(future)
```
arcondello, line 1463:
```python
        if self.config.polling_schedule.strategy == PollingStrategy.BACKOFF:
            return self._poll_using_backoff(future)
        elif self.config.polling_schedule.strategy == PollingStrategy.LONG_POLLING:
            return self._poll_using_long_polling(future)
        else:
            raise RuntimeError("unexpected polling strategy")
```
might make future development a tiny bit easier

dwave/cloud/config/models.py
```diff
@@ -101,6 +111,19 @@ class PollingSchedule(BaseModel, GetterMixin):
...
 
 
+class LongPollingSchedule(BaseModel):
+    """Problem status long polling params."""
+
+    #: Long polling strategy
+    strategy: Literal[PollingStrategy.LONG_POLLING] = PollingStrategy.LONG_POLLING
+
+    #: Maximum duration long polling connection is kept open, in seconds.
+    wait_time: Optional[int] = 30
```
arcondello, line 121: Feels slightly strange to have this as an `int` and `pause` as a `float`.

arcondello (approved):
Gave it a quick look. Will take another deeper look on a second pass, but if this is urgent is OK to merge sooner.

dwave/cloud/config/models.py
randomir: I agree, but SAPI requires `int` for `timeout` (not sure why). And `pause` is a local sleep (we'll probably remove it eventually, as I don't see many benefits/use cases).

arcondello: Might be worth a comment? :shrug:

dwave/cloud/client/base.py
randomir: Actually had it like that initially, but simplified later. But sure, can add it back, better to explicitly fail.

arcondello (approved):
Looks good on a second look as well!

PR #650, opened 2024-08-03
Implement response cache control

### LINKED ISSUE #646: Make `CachingSession` respect SAPI's `Cache-Control`/`max-age` (when released)
In #645 we use a hard-coded default value for `max-age` (15 min), and validate cache via conditional requests after that.

That's because SAPI currently doesn't implement cache control, i.e. it returns none of the `Cache-Control`, `Expires` or `Last-Modified` header fields.

Implement support for `Cache-Control` response field, as that's the most modern and most robust cache control mechanism.

randomir:
Close #646.

Until `Cache-Control` header is added to SAPI responses, we cache heuristically** for 1 hour.

** since we can't actually determine [heuristic freshness](https://www.rfc-editor.org/rfc/rfc9111#name-calculating-heuristic-fresh) as neither `Last-Modified` or `Expires` are present in SAPI responses, we default to a reasonable period for solver metadata.

### DIFF
--- a/dwave/cloud/api/client.py
+++ b/dwave/cloud/api/client.py
@@ -19,12 +19,12 @@ import hashlib
 import logging
 import os
 from collections import deque, namedtuple
-from typing import Callable, Mapping, Optional, TypedDict, Union, TYPE_CHECKING
+from typing import Callable, Mapping, Optional, Tuple, TypedDict, Union, TYPE_CHECKING
 
 import requests
 import urllib3
+import werkzeug
 from packaging.specifiers import SpecifierSet
-from werkzeug.http import parse_options_header, dump_options_header
 
 import dwave.cloud.config   # don't use `from` to break circular import (config <> api.constants)
 from dwave.cloud.api import constants, exceptions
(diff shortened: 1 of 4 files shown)


dwave/cloud/api/client.py
```diff
@@ -315,7 +315,10 @@ def configure_cache(self, cache: CacheConfig) -> None:
             logger.debug("[%s] cache disabled.", type(self).__name__)
             return
 
-        self._maxage = config['maxage']
+        if not isinstance(_maxage := config.get('maxage'), int) or _maxage < 0:
+            raise ValueError("A non-negative integer required for 'maxage'.")
+        self._maxage = _maxage
```
arcondello, line 320:
```python
       _maxage = config.get('maxage')
        if not isinstance(_maxage, int) or _maxage < 0:
            raise ValueError("A non-negative integer required for 'maxage'.")
        self._maxage = _maxage
```
Maybe this is an aesthetic preference but I found it hard to find where `_maxage` for instance came from when it's assigned in the if branch.

arcondello (approved):
LGTM

dwave/cloud/api/client.py
randomir: OK, I can update this in a follow-up PR (that I'm already working off this branch, and where I changed this exact line) to simplify conflicts on rebase.

randomir: Done in https://github.com/dwavesystems/dwave-cloud-client/pull/651.

PR #651, opened 2024-08-06
Simplify regions caching with `CachingSession`

### LINKED ISSUE #506: Make `get_regions` respect Metadata API's `Cache-Control` header's max-age directive
Currently we cache regions for 24h, which is a safe default.

However, if Metadata API response includes `Cache-Control` header, with `max-age` directive, we want to cache for up to `max-age`.

### LINKED ISSUE #647: Refactor `get_regions()` to use cache from `CachingSession`
Drop the custom implementation done for `regions`/`get_regions()`, and use the caching support provided by low-level API's `CachingSession`, as implemented in #645.

randomir:
Close #647.
Close #506.

### DIFF
--- a/dwave/cloud/api/client.py
+++ b/dwave/cloud/api/client.py
@@ -435,9 +435,10 @@ class CachingSession(VersionedAPISession):
                 logger.debug('cache hit within maxage')
                 return make_response(content)
 
-            # conditional request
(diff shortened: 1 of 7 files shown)


arcondello (approved):
LGTM

PR #656, opened 2024-08-11
Fix solver dynamic parts filtering

randomir:


dwave/cloud/client/base.py
```diff
@@ -731,7 +731,7 @@ def _fetch_solvers(self,
                     solver_id=name, filter=dynamic_fields, no_cache_=True)
 
                 # note: avg_load may not exist in solver data
-                if 'avg_load' in load:
+                if (avg_load := load.get('avg_load')) is not None:
                     solver.avg_load = load.avg_load
```
arcondello, line 735: It looks like you meant to do
```suggestion
                if (avg_load := load.get('avg_load')) is not None:
                    solver.avg_load = avg_load
```

dwave/cloud/client/base.py
```diff
@@ -749,7 +749,7 @@ def _fetch_solvers(self,
                          filter=dynamic_fields, no_cache_=True)}
 
             for solver in solvers:
-                if avg_load := loads.get(solver.id) is not None:
+                if (avg_load := loads.get(solver.id)) is not None:
```
arcondello, line 752: I am guessing that this means that we're falling back to using the id as the sort key. But setting `id` to `avg_load` might be worth a comment

arcondello (approved):

dwave/cloud/client/base.py
randomir: No, we're not -- loads is a mapping `id -> avg_load`. Unless I misunderstood you above.

arcondello: Right, but why `id` -> `avg_load`? It doesn't especially matter I just find that mapping a bit confusing out-of-context

dwave/cloud/client/base.py
randomir: Copy/paste issue form below. I actually switched to `.get` because I haven't implemented `__contains__` on the model. Will fix.

dwave/cloud/client/base.py
randomir: Because we're now fetching solvers metadata in two requests -- one for static data (cached), see `solvers` variable above; and one for dynamic data (currently not cached at all), see `loads` variable above.

Dynamic data currently is *only* `avg_load`. So we have this mapping, one `avg_load` for each solver `id`.

This loop here is going through all solvers data (static at this point), and tries to attach dynamic (`avg_load`) to each solver.

arcondello: Makes sense, thanks!

PR #670, opened 2024-09-17
Fix/simplify import dependencies

### LINKED ISSUE #669: Circular import error
In the last refactor we introduced an import error:
```
>>> from dwave.cloud.exceptions import ConfigFileError
...
ImportError: cannot import name 'ConfigFileReadError' from partially initialized module 'dwave.cloud.exceptions' (most likely due to a circular import)
```

randomir:
Make sure: (solver, computation, client) > (`dwave.cloud.exceptions`, `dwave.cloud.regions`) > `dwave.cloud.api.*` > `dwave.cloud.config.*` > `dwave.cloud.utils.*`

Fix #669.

### DIFF
--- a/dwave/cloud/exceptions.py
+++ b/dwave/cloud/exceptions.py
@@ -19,16 +19,6 @@ from dwave.cloud.api import exceptions as _api
 SAPIError = _api.RequestError
 
 
-class ConfigFileError(Exception):
-    """Base exception for all config file processing errors."""
-
-class ConfigFileReadError(ConfigFileError):
-    """Non-existing or unreadable config file specified or implied."""
-
-class ConfigFileParseError(ConfigFileError):
-    """Invalid format of config file."""
-
-
 class SolverError(_api.RequestError):
     """Generic solver-related error"""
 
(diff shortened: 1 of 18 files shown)


dwave/cloud/exceptions.py
```diff
@@ -88,3 +75,9 @@ class ProblemStructureError(InvalidProblemError):   # inherit for backwards comp
 
 class ProblemUploadError(Exception):
     """Problem multipart upload failed."""
+
+
+# for backward compatibility
+from dwave.cloud.api.exceptions import RequestTimeout
+
+from dwave.cloud.config.exceptions import *
```
thisac, line 83: Also planned for deprecation at some point?

thisac (approved):
Seems to be working well. Wasn't sure if the intention is to keep the config exceptions accessible directly from `dwave.cloud.exceptions` or, as with the constants, deprecate that usage, in which case the exception import on line 42 in `dwave.cloud.cli.py` should probably be split into two.

Otherwise, LGTM!

dwave/cloud/exceptions.py
randomir: Not any time soon, I'd say.

PR #671, opened 2024-10-03
Add Python 3.13 support

randomir:
Reworked the script on Windows to resolve and pull the latest Python version in order to support pre-releases.

Also, switched to using `python` docker images (full bookworm, ~400MB) for tests, instead of `cimg/python` ones (~800MB).

We still use the CircleCI images for jobs like build, deploy, docs, etc, but might consider moving them to standard Python images as well. The only package we would miss is `jq`, but that's required only for more advanced jobs (like some in the Ocean orb, or the windows python install).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,7 @@ jobs:
         type: string
       xcode:
         type: string
-        default: "14.2.0"
+        default: "16.0.0"
 
     macos:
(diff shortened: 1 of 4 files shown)


arcondello (approved):

thisac (approved):
LGTM

randomir:
`pyenv` from `brew` not yet updated to include 3.13.

PR #674, opened 2024-10-25
Avoid `sampleset.wait_id` overwrite with dimod 0.12.18+

randomir:
This is to support the fix to [dwave-system#540](https://github.com/dwavesystems/dwave-system/issues/540) in [dimod#1393](https://github.com/dwavesystems/dimod/pull/1393).


arcondello (approved):
LGTM!

PR #675, opened 2024-10-28
Fix live cache tests

randomir:
This fixes a couple of sporadic live cache test failures. Not really worthy of a release note.

### DIFF
--- a/dwave/cloud/api/client.py
+++ b/dwave/cloud/api/client.py
@@ -282,10 +282,6 @@ class CachingSession(VersionedAPISession):
             Force cache update, skipping time-based or etag-based validation.
     """
 
-    _cache_enabled = False
-    _maxage = None
-    _store = None
-
     class CacheConfig(TypedDict, total=False):
         enabled: bool
         maxage: float
@@ -404,6 +400,11 @@ class CachingSession(VersionedAPISession):
         refresh = kwargs.pop('refresh_', False)
         no_cache = kwargs.pop('no_cache_', False)
         default_maxage = kwargs.pop('maxage_', None)
+
+        if not self._cache_enabled or no_cache or method.lower() != 'get':
+            # completely bypass cache lookup and validation
+            return super().request(method, url, params=params, headers=headers, **kwargs)
+
         if default_maxage is None:
             default_maxage = self._maxage
         elif not isinstance(default_maxage, numbers.Real) or default_maxage < 0:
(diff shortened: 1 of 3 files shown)


dwave/cloud/api/client.py
```diff
@@ -282,10 +282,6 @@ class CachingSession(VersionedAPISession):
             Force cache update, skipping time-based or etag-based validation.
     """
 
-    _cache_enabled = False
-    _maxage = None
-    _store = None
```
thisac, line 287: Why remove these? It doesn't seem to change anything, but might cause some confusion when seeing them being accessed below. Not that it matters much, though.

randomir: Removed them exactly to reduce confusion :smile: 

The line below defines the default config, which is different from these values. Sure, there are other ways to harmonize them, but by removing them I also wanted to make sure they're not used if cache is not configured (as they're meaningless then). IMO, only `_cache_enabled` would make sense to always be available, but it's set on construction anyway...

tests/api/resources/test_solvers.py
```diff
@@ -270,23 +270,31 @@ def tearDownClass(cls):
...
+        # caching is disabled by default
+        # sapi is queried every time
+        with Solvers.from_config(validate_config_v1(config), history_size=2) as resource:
+            resource.list_solvers(filter='none,+id')
+            self.assertEqual(len(resource.session.history), 1)
+            self.assertEqual(resource.session.history[-1].response.status_code, 200)
+
+            resource.list_solvers(filter='none,+id')
+            self.assertEqual(len(resource.session.history), 2)
+            self.assertEqual(resource.session.history[-1].response.status_code, 200)
```
thisac, line 283: Just to see if I understand this correctly; is 200 is expected here twice since `history_size=2`? Would a third request return a 304 since it'd be using the cache?

tests/test_client.py
```diff
@@ -89,7 +89,7 @@ class TestTimeouts(unittest.TestCase):
     def test_request_timeout(self):
         with self.assertRaises(dwave.cloud.exceptions.RequestTimeout):
             with Client(request_timeout=0.00001, **config) as client:
-                client.get_solvers()
+                client.get_solvers(refresh=True)
```
thisac, line 92: Why is a refresh required here but not elsewhere?

dwave/cloud/api/client.py
thisac: I see your point. Why not remove their usage altogether since they seem to only used as temporary variables inside the `CachingSession` class?

thisac (approved):

dwave/cloud/api/client.py
randomir: They're not temporary variables. `_maxage` is the default maxage used if maxage is not present in server's `Cache-Control` header. `_store` is the actual key/value cached data store.

tests/test_client.py
randomir: Elsewhere we want to use the cache (to reduce runtime), and here we want to force an actual network request in order to test `request_timeout` is honored, and a `RequestTimeout` is raised if server doesn't respond in a timely manner (we use the unrealistic 10us to make sure request always fails).

dwave/cloud/api/client.py
thisac: Ok, fair enough, that actually makes sense.

PR #678, opened 2024-11-01
Add `--raw`/`-r` option to "get token" CLI commands

### LINKED ISSUE #598: Add `--raw`/`-r` option to "get token" CLI commands
Currently, to fetch a specific project's SAPI token (for use in scripting), one can write:
```
dwave leap project token --project CODE --json | jq .token -r
```
and then use it like this, for example, to list available solvers:
```
DWAVE_API_TOKEN=$(dwave leap project token --project CODE --json | jq .token -r) dwave solvers -la
```

We can simplify the JSON filtering part, though, by introducing `--raw`/`-r` option that returns raw token string:
```
$ dwave leap project token --project CODE --raw
a66...fe
```

Similarly for `dwave auth get <token-type>` command.

randomir:
Close #598.

### DIFF
--- a/dwave/cloud/cli.py
+++ b/dwave/cloud/cli.py
@@ -437,7 +437,7 @@ def standardized_output(fn):
 
         def output(fmt=None, maxlen=None, **params):
             if raw := params.pop('raw', None):
-                raw_stream.write(raw)
+                raw_stream.write(str(raw))
             info.update(params)
             if not json_output and not raw_output:
                 if fmt is not None:
--- a/dwave/cloud/cli.py
+++ b/dwave/cloud/cli.py
@@ -435,12 +435,16 @@ def standardized_output(fn):
         info = dict(datetime=now.isoformat(), timestamp=datetime_to_timestamp(now), code=0)
         raw_stream = io.StringIO()
 
-        def output(fmt, maxlen=None, **params):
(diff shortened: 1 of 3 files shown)


dwave/cloud/cli.py
```diff
@@ -419,21 +429,32 @@ def standardized_output(fn):
...
                     msg = strtrunc(msg, maxlen)
                 click.echo(msg)
 
         def flush():
-            if json_output:
+            if json_output and not raw_output:
                 click.echo(orjson.dumps(info))
+            if raw_output and raw_stream.tell():
+                raw_stream.seek(0)
+                click.echo(raw_stream.read())
```
thisac, line 457: Why use `StringIO` instead of just `click.echo(info.get("raw"))` here?

thisac: You could then also just have an elif instead
```python
if raw_output:
    ...
elif json_output:
    ...
```

randomir: The reason for using a string buffer here was to enable accumulation with `output()` calls and dumping all at the end in `flush()`, i.e. the same logic we use for JSON output. But we don't need buffering for the raw output at the moment, so sure, we can simplify the code a bit. Might go back to this implementation when/if we need buffered raw output from the CLI in the futuer.

PR #680, opened 2025-01-17
Robust client close

### LINKED ISSUE #217: Robust `Client.close()`, no client/solver/future reuse after close
After `Client.close()`:
- all workers have to be killed, immediately if user doesn't want to wait for results (which they are now too, but in a blocking manner)
- all solvers have to be invalidated, so no future submissions are possible (channel is closed),
- all futures have to be resolved (with timeout/shutdown error if `Client.close(wait=False)`)

randomir:
Close #217.

### DIFF
--- a/dwave/cloud/client/base.py
+++ b/dwave/cloud/client/base.py
@@ -65,6 +65,7 @@ from plucky import pluck
 from dwave.cloud import api
 from dwave.cloud.package_info import __packagename__, __version__
 from dwave.cloud.exceptions import *    # TODO: fix
+from dwave.cloud.exceptions import UseAfterCloseError
 from dwave.cloud.computation import Future
 from dwave.cloud.config import load_config, update_config, validate_config_v1
 from dwave.cloud.config import constants as config_constants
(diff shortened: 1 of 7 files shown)


dwave/cloud/client/base.py
```diff
@@ -439,6 +440,8 @@ def from_config(cls, config_file=None, profile=None, client=None, **kwargs):
     def __init__(self, **kwargs):
         logger.debug("Client init called with: %r", kwargs)
 
+        self._initialized = False
```
arcondello, line 443: This is so that if there is an error in the next block we're still left in a valid state? Is it possible to detect whether the client is currently initialized via other attribute(s) saved on the object?

arcondello (approved):
Seems reasonable to me. I suppose something like

```python
def require_initialized(method):
    def _method(obj, *args, **kwargs):
        if not obj._initialized:
            raise UseAfterCloseError(f"{method.__name__} cannot be called after client has been closed")
        return method(obj, *args, **kwargs)
    return _method
```

might be a bit more uniform. But probably also overkill for what you're going for?

dwave/cloud/client/base.py
randomir: I wanted to make sure this flag is available in worker threads that are started before object finishes initializing. Ended up not needing the flag in workers, but still it's probably safer this way.

PR #682, opened 2025-01-28
Thread-safe client close

randomir:
A follow-up to #680, in which we make sure client use is disabled also during _closing_, not just after the client has been closed.

We also make sure `close()` is thread-safe wrt attempted usage while closing.

~TODO: consider converting `Client.solvers_session` to a ctxmgr, or making this session ephemeral in another way.~
:arrow_right:  Keeping `Client.solvers_session` active during client lifespan, as it saves ~3ms on init due to disk cache access.

### DIFF
--- a/dwave/cloud/client/base.py
+++ b/dwave/cloud/client/base.py
@@ -86,9 +86,9 @@ __all__ = ['Client']
 logger = logging.getLogger(__name__)
 
 
-def _require_initialized(method):
+def _ensure_active(method):
     def _method(obj, *args, **kwargs):
-        if not getattr(obj, '_initialized', False):
+        if getattr(obj, '_closed', False):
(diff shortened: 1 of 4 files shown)


arcondello (approved):
LGTM!

I suppose if you also wanted a context manager you could do something with a classmethod. Maybe
```python
class Client:
    @contextlib.contextmanager
    @classmethod
    def as_ctx(cls, *args, **kwargs):
        client = cls(*args, **kwargs)
        yield client
        client.close()
```
might be less performant but convenient in some cases?

randomir:
@arcondello, `Client` already is a context manager (see `__enter__` and `__exit__` methods).

I considered converting `solvers_session` property to a context manager, but decided not to do it for now to save some init time.

PR #683, opened 2025-02-01
Add payload compression support to `DWaveAPIClient`

### LINKED ISSUE #654: Add support for payload compression to `DWaveAPIClient`
To enable QPU problem data upload compression on the low-level API.

randomir:
WIP.

Close #654.

### DIFF
--- a/dwave/cloud/api/client.py
+++ b/dwave/cloud/api/client.py
@@ -20,8 +20,10 @@ import logging
 import numbers
 import os
 import warnings
+import zlib
 from collections import deque, namedtuple, abc
-from typing import Optional, TypedDict, Union, TYPE_CHECKING
+from collections.abc import Iterable
+from typing import IO, Optional, TypedDict, Union, TYPE_CHECKING
 
 import orjson
 import requests
@@ -131,9 +133,81 @@ class LoggingSession(BaseUrlSession):
         return response
 
 
-class VersionedAPISession(LoggingSession):
+class PayloadCompressingSession(LoggingSession):
     """A `requests.Session` subclass (technically, further specialized
-    :class:`.LoggingSession`) that enforces conformance of API response
+    :class:`.LoggingSession`) that adds support for payload compression on the
+    fly.
+
+    Args:
+        compress (bool, default=False):
+            Preemptively compress all payloads sent, even on PUT and POST requests.
+
+    """
+
+    _compress: bool = False
+
+    def __init__(self, compress: bool = False, **kwargs):
+        self._compress = compress
+        super().__init__(**kwargs)
+
+    @staticmethod
+    def _iter_compressed(data: Union[bytes, IO, Iterable[bytes]],
+                         chunk_size: Optional[int] = None):
+        if chunk_size is None:
+            # match zlib window size
+            chunk_size = 2**15
+
+        if isinstance(data, bytes):
+            data = io.BytesIO(data)
+
+        if hasattr(data, 'read'):
+            def get_chunks(f, chunk_size):
+                while chunk := f.read(chunk_size):
+                    yield chunk
+            chunks = get_chunks(data, chunk_size)
+
+        else:
+            # assume data is iterable
+            chunks = data
+
+        zbuf = zlib.compressobj()
+
+        for chunk in chunks:
+            if isinstance(chunk, str):
+                chunk = chunk.encode("utf-8")
+            if c := zbuf.compress(chunk):
+                yield c
+
+        yield zbuf.flush()
+
+    def request(self, *args, **kwargs):
+        compress = kwargs.pop('compress', self._compress)
+
+        # enabled compression of current request (assumes Session is not thread-safe)
+        old = self._compress
+        try:
+            self._compress = compress
+            return super().request(*args, **kwargs)
+        finally:
+            self._compress = old
+
+    def send(self, request: requests.PreparedRequest, **kwargs) -> requests.Response:
+        # Note: compress body on `send()` (rather than on `request()`) as by now
+        # json, form, and multipart data are all converted to bytes. We only need
+        # to handle bytes, files and iterators at this point.
+        if self._compress and request.body:
+            request.body = self._iter_compressed(request.body)
+            # update the prepared request for chunked upload
+            request.headers['Content-Encoding'] = 'deflate'
+            request.headers['Transfer-Encoding'] = 'chunked'
+            request.headers.pop('Content-Length', None)
+
+        return super().send(request, **kwargs)
+
+
+class VersionedAPISession(PayloadCompressingSession):
+    """A `requests.Session` subclass (technically, further specialized
+    :class:`.PayloadCompressingSession`) that enforces conformance of API response
     version with supported version range(s).
 
     Response format version is requested via `Accept` header field, and
--- a/tests/api/test_api_clients.py
+++ b/tests/api/test_api_clients.py
@@ -16,6 +16,7 @@ import json
 import time
 import uuid
 import unittest
+import zlib
 
 import diskcache
 import requests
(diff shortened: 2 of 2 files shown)


dwave/cloud/api/client.py
```diff
@@ -20,8 +20,10 @@
 import numbers
 import os
 import warnings
+import zlib
 from collections import deque, namedtuple, abc
-from typing import Optional, TypedDict, Union, TYPE_CHECKING
+from collections.abc import Iterable
+from typing import IO, Optional, TypedDict, Union, TYPE_CHECKING
```
JoelPasvolsky, line 26:
```suggestion
from typing import IO, Optional, TypedDict, TYPE_CHECKING, Union
```

dwave/cloud/api/client.py
```diff
@@ -131,9 +133,108 @@ def request(self, method: str, *args, **kwargs):
         return response
 
 
-class VersionedAPISession(LoggingSession):
+class PayloadCompressingSession(LoggingSession):
+    """A :class:`requests.Session` subclass (technically, a further specialized
+    :class:`.LoggingSession`) that adds support for payload compression on the
```
JoelPasvolsky, line 138: These two classes do not create links because Requests is not part of our intersphinx map. Consider making links to the docs for these in the `requests.Session <https://requests.readthedocs.io/en/latest/api/#requests.Session>`_ format

dwave/cloud/api/client.py
```diff
@@ -131,9 +133,108 @@ def request(self, method: str, *args, **kwargs):
...
+    fly.
+
+    Args:
+        compress (bool, default=False):
+            Preemptively compress HTTP request body, even on PUT and POST requests.
+            When compression is enabled, the payload is always compressed in 32KiB
```
JoelPasvolsky, line 144:
```suggestion
            When compression is enabled, the payload is always compressed in 32KB
```

dwave/cloud/api/client.py
```diff
@@ -131,9 +133,108 @@ def request(self, method: str, *args, **kwargs):
...
 
-class VersionedAPISession(LoggingSession):
+class PayloadCompressingSession(LoggingSession):
+    """A :class:`requests.Session` subclass (technically, a further specialized
+    :class:`.LoggingSession`) that adds support for payload compression on the
+    fly.
+
+    Args:
+        compress (bool, default=False):
+            Preemptively compress HTTP request body, even on PUT and POST requests.
```
JoelPasvolsky, line 143: "even on PUT and POST requests" --> to me it's unclear for what other requests the expected behavior is compression that these are now added to

dwave/cloud/api/client.py
```diff
@@ -131,9 +133,108 @@ def request(self, method: str, *args, **kwargs):
...
+            When compression is enabled, the payload is always compressed in 32KiB
+            chunks and streamed to minimize memory overhead.
+
+    .. note::
+        With compression enabled, the request body is transferred using chunked
+        encoding (``Transfer-Encoding: chunked``) and content is compressed using
+        deflate compression algorithm (``Content-Encoding: deflate``).
```
JoelPasvolsky, line 150:
```suggestion
        the `DEFLATE <http://www.faqs.org/rfcs/rfc1951.html>`_ compression algorithm
        (``Content-Encoding: deflate``).
```

dwave/cloud/api/client.py
```diff
@@ -131,9 +133,108 @@ def request(self, method: str, *args, **kwargs):
...
+    def __init__(self, compress: bool = False, **kwargs):
+        self._compress = compress
+        super().__init__(**kwargs)
+
+    def set_payload_compress(self, compress: bool = True) -> bool:
+        """Modify payload compression on upload for current session.
+
+        Args:
+            compress (bool, default=True):
+                Enable request body compression with ``deflate`` on upload
```
JoelPasvolsky, line 168:
```suggestion
                Enable request body compression with DEFLATE on upload
```

if you mean here to specify the algorithm. If you mean to say how the header is modified: ``Content-Encoding: deflate``

dwave/cloud/api/client.py
```diff
@@ -131,9 +133,108 @@ def request(self, method: str, *args, **kwargs):
...
+                requests like ``POST`` and ``PUT``. Make sure the server supports
```
JoelPasvolsky, line 169: Which server is this referring to? The user's computer?

dwave/cloud/api/client.py
```diff
@@ -131,9 +133,108 @@ def request(self, method: str, *args, **kwargs):
...
+            # update the prepared request for chunked upload
+            request.headers['Content-Encoding'] = 'deflate'
+            request.headers['Transfer-Encoding'] = 'chunked'
+            request.headers.pop('Content-Length', None)
+
+        return super().send(request, **kwargs)
+
+
+class VersionedAPISession(PayloadCompressingSession):
     """A `requests.Session` subclass (technically, further specialized
```
JoelPasvolsky, line 236: see comment for ``PayloadCompressingSession`` class on the class

dwave/cloud/api/resources.py
```diff
@@ -58,6 +58,31 @@ def wrapper(obj: 'ResourceBase', *args, **kwargs):
         return wrapper
 
 
+class compress_if:
+    """Decorate :class:`ResourceBase` methods to enable compression on outbound
```
JoelPasvolsky, line 62: :class:`ResourceBase` needs full path

dwave/cloud/api/client.py
```diff
@@ -131,9 +133,108 @@ def request(self, method: str, *args, **kwargs):
...
+            def get_chunks(f, chunk_size):
+                while chunk := f.read(chunk_size):
+                    yield chunk
+            chunks = get_chunks(data, chunk_size)
+
+        else:
+            # assume data is iterable
+            chunks = data
+
+        zbuf = zlib.compressobj()
```
arcondello, line 200: I believe this uses default compression. I have no reason to think that's a bad level, but do we want to make the configurable? Might be too low level of a knob to expose :thinking:

arcondello (approved):

dwave/cloud/api/client.py
randomir: Note that the whole `dwave.cloud.api.*` subspace is not added to docs yet (it's considered private), but point taken - I'll add Requests to the intersphinx map.

dwave/cloud/api/client.py
randomir: KiB is the correct technical term here, and I'm sticking to it. :slightly_smiling_face:

dwave/cloud/api/client.py
randomir: The server this client (session) is talking to. Typically, that's SAPI, but not in general.

dwave/cloud/api/client.py
randomir: Fair point; yes, we use the default compression level (`-1` / `Z_DEFAULT_COMPRESSION`).
>  `Z_DEFAULT_COMPRESSION` represents a default compromise between speed and compression (currently equivalent to level 6).

It's a very low-level knob to expose (on an already pretty low-level interface), and I can't think of a practical use case for this feature. However, if anyone asks for it, it's trivial to add it.. so, until then..

randomir: Thinking about this, making the compression algorithm selectable (adding Brotli or Zstd) would be nice, but probably an overkill with questionable benefits.

dwave/cloud/api/client.py
JoelPasvolsky: Okay, I added it (and ``urllib3``) to the **SDK's** intersphinx map, where it will actually do something useful😁.
(In the unification PR.)

randomir: Thanks! I was planning to do that prior to releasing 8.2.

dwave/cloud/api/client.py
randomir: Did some zlib level benchmarks on a random P16 problem (float biases between -1 and 1). Uncompressed encoded QPU data is 492039 bytes. And levels 5-9 result in the same compression ratio:
```
0: 492090
1: 376840
2: 376748
3: 376739
4: 370557
5: 370552
6: 370552
7: 370552
8: 370552
9: 370552
```

arcondello: Totally agree it's too low level to add eagerly

PR #694, opened 2025-04-25
Switch from pkgutil to native namespace package

randomir:


.circleci/config.yml
```diff
@@ -61,10 +61,6 @@ jobs:
               pip uninstall dimod -y
             fi
 
-      - run: &install-package
-          name: Install package
-          command: env/bin/pip install .
```
thisac, line 66: Likely missing something, but why aren't these needed anymore?

thisac (approved):

arcondello (approved):
Amazing!

PR #718, opened 2025-08-22
Fix live config tests by isolating env

randomir:
Otherwise, later tests that depend on the env (like token in env) will fail.


tests/test_config.py
```diff
@@ -76,9 +77,11 @@ def test_config_load_from_file(self):
     def setUp(self):
         # clear `config_load`-relevant environment variables before testing, so
         # we only need to patch the ones that we are currently testing
```
thisac, line 79: I could've swore this said `empty=True` when I first looked and just when I was about to comment on it it switched. Now I don't know what I saw anymore. 😅

thisac (approved):

tests/test_config.py
randomir: I said empty at first, but that was a mistake on my part. Forgot later tests actually depend on env vars like `HOME` on linux or `SYSTEMDRIVE` on windows...

PR #719, opened 2025-08-25
Add `dwave cache` CLI

randomir:


dwave/cloud/utils/decorators.py
```diff
@@ -287,7 +288,7 @@ def ondisk(cls, **kwargs):
         # defer to break circular imports
         from dwave.cloud.config import get_cache_dir
 
-        directory = kwargs.pop('directory', get_cache_dir())
+        directory = kwargs.pop('directory', os.path.join(get_cache_dir(), 'func'))
```
thisac, line 291: Why this change?

thisac (approved):
LGTM

dwave/cloud/utils/decorators.py
randomir: But the main reason to do it here and now -- is to separate the two cache locations, so we can inspect and delete them in isolation easier.

PR #720, opened 2025-09-03
Enable wait on client close

### LINKED ISSUE #705: UseAfterCloseError error message could be more helpful
After the introduction of the ``close()`` method and recommendation to use a context manager for samplers, a typical example looks like this:

```
>>> with DWaveSampler() as sampler:
...    sampler.sample_ising(...)
```

I suspect users testing simple submissions will use the above and only then use the returned value (outside the context manager) because that typically works well, but in this case the use of a Future for the asynchronous submission will cause an error with an error message that will be unclear to most users:

```
>>> with DWaveSampler() as sampler:
...    sampler.sample_ising(...)
>>> print(sampleset)
...
UseAfterCloseError: _poll cannot be called after client has been closed
```

What about adding another line to that error message for the typical use case of exiting a context manager without resolving the Future? That seems to me preferable than pointing out that the sampleset needs to be resolved before exiting the context manager is all such examples.

### LINKED ISSUE #712: Institute previous client close behavior to wait for all jobs to finish
The following change in `close()`, made as part of #682:

https://github.com/dwavesystems/dwave-cloud-client/blob/a7720dd8fe6921a5fae29e20c97532b496c0603d/dwave/cloud/client/base.py#L633-L639

is preventing **new** problem status polling/result download jobs (alongside with problem submission/cancel) after the `close()` was called.

This makes the close operation more efficient and faster, as that close (or sampler context exit) could cause potentially very long wait times (think of long hybrid jobs).

But at the same time it causes unexpected behavior, like reported in #705 -- where the sampleset future is marked failed (with polling after close forbidden error), only to surface when that future is resolved.

Still not sure what's a better default behavior, but we can at least add an option, `wait`, to `close()` to prevent exit before all remote jobs finish.

Samplers could accept e.g. `wait_on_close` in their init, they can then pass to client close.

Another option would be to add a `timeout` parameter, but implementing that seems tricky.

randomir:
Close #712.
Fix #705.

### DIFF
--- a/dwave/cloud/client/base.py
+++ b/dwave/cloud/client/base.py
@@ -447,7 +447,11 @@ class Client(object):
         logger.debug("Creating %s.Client() with: %r", _client, config)
         return _clients[_client](**config)
 
-    class _CountDownLatch:
+    class _WaitableCounter:
+        """A thread-safe counter, with an ability to block until the count
+        reaches zero (use :meth:`.wait`).
+        """
+
         def __init__(self, value: int = 0):
             self._cond = threading.Condition()
             self._value = value
@@ -478,7 +482,7 @@ class Client(object):
         self._closed = False
         self._closing = False
         self._close_lock = threading.Lock()
-        self._jobs = self._CountDownLatch()
+        self._jobs = self._WaitableCounter()
 
         # derive instance-level defaults from class defaults and init defaults
         self.defaults = copy.deepcopy(self.DEFAULTS)
@@ -639,7 +643,7 @@ class Client(object):
                                 self._poll_workers, self._load_workers):
                 worker.join()
 
-    def close(self, wait: bool = DEFAULT_WAIT_ON_CLOSE):
+    def close(self, wait: Optional[bool] = None):
         """Perform a clean shutdown.
 
         Waits for all the currently scheduled work to finish, kills the workers,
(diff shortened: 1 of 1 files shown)


dwave/cloud/client/base.py
```diff
@@ -445,12 +447,38 @@ def from_config(cls, config_file=None, profile=None, client=None, **kwargs):
         logger.debug("Creating %s.Client() with: %r", _client, config)
         return _clients[_client](**config)
 
+    class _CountDownLatch:
```
arcondello, line 450: I found this class a bit confusingly named. It looks like we currently always instantiate it with `value=0` and then inc/decrement it. I would have expected based on its name that we instantiate it with a positive `value` and then only decrement. IMO a short comment/docstring would help.

dwave/cloud/client/base.py
```diff
@@ -609,12 +639,22 @@ def _shutdown_threads(self, wait: bool = True):
                                 self._poll_workers, self._load_workers):
                 worker.join()
 
-    def close(self):
+    def close(self, wait: bool = DEFAULT_WAIT_ON_CLOSE):
```
arcondello, line 642: If we change `Client.DEFAULT_WAIT_ON_CLOSE` it won't change the behavior of this method. So we should either do `Client._DEFAULT_WAIT_ON_CLOSE` or make this a `None` etc etc.

arcondello (approved):
Seems more intuitive to me!

dwave/cloud/client/base.py
randomir: I agree the name is confusing, will update it to something better.

PR #722, opened 2025-09-05
Add support for sapi deprecation messages

### LINKED ISSUE #721: Add support for Leap deprecation messages
Leap/SAPI recently added support for announcing deprecations (of solvers, features, APIs, etc) via response header fields.

We would like to parse those messages and raise appropriate warnings, most likely [`DeprecationWarning`s](https://docs.python.org/3/library/exceptions.html#DeprecationWarning).

randomir:
Close #721.

### DIFF
--- a/dwave/cloud/api/client.py
+++ b/dwave/cloud/api/client.py
@@ -622,13 +622,25 @@ class DeprecationAwareSessionMixin:
                     Enable deprecation message logging (using ``WARNING`` log level).
 
                 * ``warn`` (bool):
-                    Raise each deprecation message using
-                    :class:`~dwave.cloud.api.exceptions.ResourceDeprecationWarning`
+                    Raise each deprecation message using :exc:`DeprecationWarning`,
+                    or :class:`~dwave.cloud.api.exceptions.ResourceDeprecationWarning`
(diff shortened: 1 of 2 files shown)


dwave/cloud/api/exceptions.py
```diff
@@ -59,3 +59,7 @@ class InternalServerError(RequestError):
 
 class RequestTimeout(RequestError):
     """API request timed out"""
+
+
+class ResourceDeprecationWarning(FutureWarning):
```
randomir, line 64: By subclassing `FutureWarning`, we're making sure deprecation warnings are not silenced by default. Might revisit this later.

tests/api/test_api_clients.py
```diff
@@ -12,22 +12,24 @@
...
 
+import datetime
 import json
 import time
 import uuid
 import unittest
 import zlib
 
 import diskcache
+import http_sf
```
arcondello, line 23: I am sure you thought about it but `http_sf` does not seem like it has a tonne of adoption. I think there is some risk that it gets abandoned. Can always adjust later but :shrug:

arcondello (approved):

PR #728, opened 2025-10-06
Implement adaptive `retrieve_answer()` to handle partial v3 solver representions

### LINKED ISSUE #727: `Client.retrieve_answer()` fails when a problem was submitted using solver v2 representation
When problem is submitted using version 2 format (Ocean 8, `dwave-cloud-client<0.14`), only solver's name is stored in problem status structure (field `solver` is a string solver name).

Then, when problem status (with answer) is downloaded (from `problems/<id>/`), using version 3 format (Ocean 9, `dwave-cloud-client>=0.14`), the `solver` field looks consistent with the v3 format (it's a dict), but it contains only the `name`. `graph_id` is missing! Since v3 requires name+graph_id for QPU solvers, when client tries to load solver with identity that lacks the `graph_id`, it fails with solver not found.

Until a fix is implemented in SAPI to return a valid solver identity (for problems submitted using v2 format), we can implement a workaround in cloud-client -- during problem status/answer retrieval, if `graph_id` is missing, we can do a "partial" solver match using just the name, accompanied by a user warning.

randomir:
Close #727.

### DIFF
--- a/dwave/cloud/client/base.py
+++ b/dwave/cloud/client/base.py
@@ -1462,7 +1462,23 @@ class Client(object):
                     # An alternative to making this call here would be to pass
                     # self in with the message
                     if future.solver is None:
-                        future.solver = self.get_solver(identity=message['solver'])
+                        # handle problems with v2 solver representation
+                        # (prior to `graph_id` introduction)
+                        # see: https://github.com/dwavesystems/dwave-cloud-client/issues/727
+                        # TODO: remove after a server-side fix
+                        solver = message['solver']
+                        if isinstance(solver, dict):
+                            if 'version' in solver:
+                                # full identity match for valid v3 response
+                                future.solver = self.get_solver(identity=solver)
+                            elif 'name' in solver:
+                                # partial solver match; valid during v2 deprecation period
+                                future.solver = self.get_solver(name=solver['name'])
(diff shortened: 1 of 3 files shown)


charleswhchan (approved):

dwave/cloud/client/base.py
```diff
@@ -1462,7 +1462,23 @@ def _handle_problem_status(self, message, future):
                     # An alternative to making this call here would be to pass
                     # self in with the message
                     if future.solver is None:
-                        future.solver = self.get_solver(identity=message['solver'])
+                        # handle problems with v2 solver representation
+                        # (prior to `graph_id` introduction)
+                        # see: https://github.com/dwavesystems/dwave-cloud-client/issues/727
+                        # TODO: remove after a server-side fix
+                        solver = message['solver']
+                        if isinstance(solver, dict):
```
arcondello, line 1470: Do we want `collections.abc.Mapping`? Probably overly general...

arcondello (approved):

dwave/cloud/client/base.py
randomir: Not needed here, as this message always comes from JSON, so types are pretty limited.

PR #734, opened 2025-11-01
Add cache control through config interface

### LINKED ISSUE #614: Implement cache control via environment variable(s)
Similarly to config file location environment variable, `DWAVE_CONFIG_FILE`, add an option to **control cache directory** location (for region and solver metadata cache, etc) from the environment.

For example, `DWAVE_CACHE_HOME`.

An option to **disable disk caching** could be implemented either via a separate variable, or as a special-meaning value, such as `DWAVE_CACHE_HOME=disabled`.

While implementing this, have in mind we'll eventually want to have a shared Ocean cache control/interface, see https://github.com/dwavesystems/minorminer/issues/242.

Another control dimension would be cache type: memory (`dwave.cloud.utils.cached`), disk (`dwave.cloud.utils.cached.ondisk` or `diskcache`), memcached?, redis?, etc.

randomir:
Close #614.

### DIFF
--- a/dwave/cloud/cli.py
+++ b/dwave/cloud/cli.py
@@ -49,7 +49,7 @@ from dwave.cloud.config import (
     load_profile_from_files, load_config_from_files, load_config, get_default_config,
     get_configfile_path, get_default_configfile_path, get_configfile_paths,
     get_cache_dir)
-from dwave.cloud.config.models import validate_config_v1
+from dwave.cloud.config.models import validate_config_v1, ClientConfig
 from dwave.cloud.regions import get_regions
 from dwave.cloud.auth.flows import LeapAuthFlow, OAuthError
 from dwave.cloud.auth.creds import _get_creds_paths, Credentials
@@ -1212,7 +1212,7 @@ def cache():
     """Interact with Ocean cache."""
 
 
-def _get_cache_info() -> dict[str, list[dict]]:
+def _get_cache_info(config: ClientConfig) -> dict[str, list[dict]]:
     def _stats(cache: diskcache.Cache) -> dict:
         return {
             'count': len(cache),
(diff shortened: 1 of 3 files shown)


dwave/cloud/api/client.py
```diff
@@ -801,14 +817,14 @@ class DefaultSession(VersionedAPISessionMixin,
...
-        'cache': dict(enabled=False),       # type: CachingSession.CacheConfig
+        'cache': dict(enabled=False),       # type: CachingSession.ExtendedCacheConfig
 
         # api deprecation message handling, see :class:`DeprecationAwareSessionMixin`
         'on_deprecation': dict(log=True, warn=True, store=True),
     }
 
     # client instance config, populated on init from kwargs overriding DEFAULTS
-    config = None
+    config: dict
```
thisac: Just confirming that there's no case where `DWaveAPIClient.config` might be called (on the class), expecting a `None` to return? With this change an `AttributeError` would be raised I believe.

dwave/cloud/config/models.py
```diff
@@ -125,6 +127,42 @@ class LongPollingSchedule(BaseModel):
...
+
+        # expand home sentinels (override enabled)
+        if self.home is not None:
+            home = self.home.lower()
+            if home in ['off', 'disabled']:
+                self.enabled = False
+                self.home = None
+            elif home == 'default':
+                self.enabled = True
+                self.home = get_cache_dir()
```
thisac: Should caching always be enabled when "home" is set _even_ if explicitly set to false?

randomir: Caching is enabled only if a sentinel value `default` is used for `home`, so that a user can easily enable it from the env, even if it's explicitly disabled in the config file.

dwave/cloud/api/client.py
randomir: Nope, that's why I simplified it here.

PR #736, opened 2025-11-05
Implement adaptive `retrieve_answer()` to handle incorrect v3 solver representions

randomir:
Handle answer retrieval or problems with `null` for solver version.

See also #727.


arcondello (commented):
LGTM

PR #744, opened 2026-03-27
Replace pickle with json+raw for response caching and auth creds

### LINKED ISSUE #742: Replace pickle with custom serialization in `CachingSessionMixin`
`CachingSessionMixin` defaults to Python pickle for serializing API responses to an on-disk cache (SQLite database via `diskcache` library).

For simpler Metadata API caching we [used to use JSON](https://github.com/dwavesystems/dwave-cloud-client/pull/559), but when implementing a general response caching (see #645) we defaulted to pickle to be able to store non-JSON-serializable types like `bytes`, given that the security risk was minimal in typical use cases (single-user systems, and/or containerized workflows).

Risk mitigation in other use cases was further enhanced with the introduction of a cache control interface in #734, allowing users to completely disable response caching, or to control the cache file location -- enabling easier file access control.

Nevertheless, to completely remove the risk of a "pickle attack", we should implement a custom serializer that can handle raw responses (typically `bytes`).

randomir:
Close #742.

### DIFF
--- a/dwave/cloud/api/client.py
+++ b/dwave/cloud/api/client.py
@@ -421,7 +421,7 @@ class CachingSessionMixin:
 
         # defer import and construction until needed
         import diskcache
-        return diskcache.Cache(directory=directory, **kwargs)
+        return diskcache.Cache(disk=diskcache.JSONDisk, directory=directory, **kwargs)
 
     # default cache config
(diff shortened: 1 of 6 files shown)


thisac (approved):
LGTM

PR #745, opened 2026-05-13
Add preliminary QCDL support

randomir:
Example:
```python
import orjson
from dwave.cloud import Client

client = Client.from_config()
solver = client.get_solver(supported_problem_types__contains='qcdl')

qcdl = {...}
params = dict(shots=3, time_limit=600, noisy=False)
# quick test with: `qcdl, params = solver.minimal_job`

response = solver.sample_qcdl(qcdl, **params)
answer = orjson.loads(response.answer_data.read())
```

Alternatively, to upload the circuit first, replace the `sample_qcdl` line with:
```python
circuit_id = solver.upload_qcdl(qcdl).result()
response = solver.sample_qcdl(circuit_id, **params)
```

**AI Generation Disclosure**

No AI used.


dwave/cloud/solver.py
```diff
@@ -1603,10 +1603,71 @@ def estimate_qpu_access_time(self,
...
+
+        data (`dict`):
+            Data from the server describing this solver.
+    """
+
+    _handled_problem_types = {"qcdl"}
+    _handled_encoding_formats = {"binary-ref"}
+
+    def _encode_problem_for_upload(self, qcdl, **kwargs):
+        return orjson.dumps(qcdl)
```
arcondello, line 1624: I wonder if a `try`/`except` here might help some users debug in the future.

OTOH we don't have evidence that this is a common failure path, so until we do probably better to leave it alone. Ok, I talked myself out of it, but leaving the comment for posterity 😄

arcondello (approved):
I have always loved this design. No more and no less than the amount of information needed to support the new problem type.

dwave/cloud/solver.py
randomir: Once we integrate QCDL generation into Ocean, we'll be able to do better checks here. Or even defer serialization completely. Just like we do with NL (and require dwave-optimization) or BQM/CQM/DQM (and require dimod).

randomir:
Due to a lot of unrelated changes required, I'll implement ping support for the QCDL solver in a follow-up PR.

PR #746, opened 2026-05-13
Upgrade http-sf to fix date parsing

randomir:
Older versions of http-sf didn't correctly parse input dates. Namely, naive datetime was produced instead of timezone-aware. Note the input date is coded as unix epoch timestamp, so it's always in UTC -- that was not reflected in the parsed result.

See also: https://github.com/mnot/http-sf/pull/21.


arcondello (approved):
good catch!

PR #747, opened 2026-05-21
Define minimal problem per solver class and simplify solver pinging

randomir:
In this PR we:
- redefine base solvers and their subclasses a bit; now `minimal_problem`, `sample_problem` and `upload_problem` are universally defined for all solvers.
- add support for pinging any existing solver, without the need to manually specify required parameters

### DIFF
--- a/dwave/cloud/solver.py
+++ b/dwave/cloud/solver.py
@@ -217,6 +217,23 @@ class BaseSolver:
         else:
             raise ValueError("Don't know how to decode %r answer format" % fmt)
 
+    # Sampling methods
+
+    def sample_problem(self,
+                       problem: Any,
+                       *,
+                       label: str | None = None,
+                       problem_type: str | None = None,
+                       upload_params: dict | None = None,
+                       **sample_params: Any,
+                       ) -> Future:
+        """Sample from the specified problem."""
+        raise NotImplementedError
+
+    def upload_problem(self, problem: Any, **kwargs: Any):
+        """Encode and upload the specified problem."""
+        raise NotImplementedError
+
     # Derived properties
 
     @property
@@ -318,7 +335,7 @@ class QuadraticUnstructuredSolverMixin:
             import dimod
         except ImportError: # pragma: no cover
             raise RuntimeError("Can't use unstructured quadratic solver without dimod. "
-                               "Re-install the library with bqm support.")
+                               "Re-install the library with 'bqm' support.")
 
         bqm = dimod.BinaryQuadraticModel.from_ising(linear, quadratic, offset)
         return self.sample_bqm(bqm, label=label, **params)
@@ -353,7 +370,7 @@ class QuadraticUnstructuredSolverMixin:
             import dimod
         except ImportError: # pragma: no cover
             raise RuntimeError("Can't use unstructured quadratic solver without dimod. "
-                               "Re-install the library with bqm support.")
+                               "Re-install the library with 'bqm' support.")
 
         bqm = dimod.BinaryQuadraticModel.from_qubo(qubo, offset)
         return self.sample_bqm(bqm, label=label, **params)
@@ -1222,6 +1239,50 @@ class StructuredSolver(BaseSolver):
 
         return self._sample(problem_type, linear, quadratic, bqm.offset,
                             params, label=label, undirected_biases=True)
+
+    def sample_problem(self,
+                       problem: tuple[list | dict[int, float],
+                                      dict[tuple[int, int], float],
+                                      float],
+                       *,
+                       problem_type: Literal['ising', 'qubo'] = 'ising',
+                       undirected_biases: bool = False,
+                       label: str | None = None,
+                       upload_params: dict | None = None,
+                       **sample_params: Any,
+                       ) -> Future:
+        """Sample from the specified problem.
+
+        Args:
+            problem:
+                Tuple with linear/quadratic terms of the model and the offset.
+
+            problem_type:
+                Problem type, one of the handled problem types by the solver.
+
+            undirected_biases:
+                Are (quadratic) biases specified on undirected edges? For
+                triangular or symmetric matrix of quadratic biases set it to
+                ``True``.
+
+            label:
+                Problem label you can optionally tag submissions with for ease
+                of identification.
+
+            upload_params:
+                Optional upload/encode parameters, not used by structured solvers.
+
+            **sample_params:
+                Sampling parameters, solver-specific.
+
+        Returns:
+            Response in a future.
+
+        """
+        linear, quadratic, offset = problem
+        return self._sample(type_=problem_type, linear=linear, quadratic=quadratic,
+                            offset=offset, undirected_biases=undirected_biases,
+                            label=label, params=sample_params)
 
     @dispatches_events('sample')
     def _sample(self, type_, linear, quadratic, offset, params,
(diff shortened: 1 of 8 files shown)


dwave/cloud/solver.py
```diff
@@ -222,16 +218,31 @@ def decode_response(self, msg, answer_data: Optional[io.IOBase] = None):
             raise ValueError("Don't know how to decode %r answer format" % fmt)
 
     # Sampling methods
-    def sample_ising(self, linear, quadratic, **params):
-        raise NotImplementedError
 
-    def sample_qubo(self, qubo, **params):
+    def sample_problem(self,
```
randomir, line 222: Not sure "sample" is the right verb here, long term.

dwave/cloud/solver.py
```diff
@@ -222,16 +218,31 @@ def decode_response(self, msg, answer_data: Optional[io.IOBase] = None):
...
         raise NotImplementedError
 
-    def sample_bqm(self, bqm, **params):
+    def upload_problem(self, problem: Any, **kwargs: Any):
+        """Encode and upload the specified problem."""
         raise NotImplementedError
 
-    def upload_bqm(self, bqm):
+    @property
+    def minimal_problem(self) -> tuple[Any, dict]:
```
randomir, line 238: Similarly, not sure if "job" instead of problem would be better. I chose "problem" for consistency across Leap, SAPI and Ocean. 🤷

dwave/cloud/solver.py
arcondello: Agreed. Probably need a much more generic verb like "run".

dwave/cloud/solver.py
```diff
@@ -564,6 +583,28 @@ class BQMSolver(BaseUnstructuredSolver):
...
+    def minimal_problem(self) -> tuple['dimod.BinaryQuadraticModel', dict]:
+        """A small :class:`~dimod.BinaryQuadraticModel` accepted by the
+        :meth:`.sample_problem` method, together with a set of required
+        parameter values.
+        """
+        try:
+            from dimod import BinaryQuadraticModel
+        except ImportError: # pragma: no cover
+            raise RuntimeError(
+                "dimod package with support for DiscreteQuadraticModel required."
```
arcondello: copy and paste error I think

dwave/cloud/solver.py
```diff
@@ -564,6 +583,28 @@ class BQMSolver(BaseUnstructuredSolver):
...
+                "Re-install the library with 'dqm' support.")
+
+        bqm = BinaryQuadraticModel.from_ising({0: 1}, {})
```
arcondello: Do we maybe want like a single antiferromagnetic coupling? Not sure it matters but somehow a single linear bias feels somehow *too* minimal 😆 

I suppose this way lies madness.

dwave/cloud/solver.py
```diff
@@ -818,6 +908,27 @@ def _encode_problem_for_upload(self,
...
+        except ImportError: # pragma: no cover
+            raise RuntimeError("Can't sample from nonlinear model without dwave-optimization. "
+                               "Re-install the library with 'nlm' support.")
+
+        model = Model()
+        x = model.list(10)
+        model.minimize(x.sum())
+
+        # note: minimum time limit is not exposed via properties, but value >0 is required
+        params = dict(time_limit=1)
```
arcondello, line 928: Is it not? https://github.com/dwavesystems/dwave-system/blob/ec8cbdf52f36f4a7f9ef6249a1325c5ef290a373/dwave/system/samplers/leap_hybrid_sampler.py#L1063 It's supposed to be 5s for the "minimum default" though it will accept any nonnegative value as you note.

arcondello (approved):
Seems reasonable to me and a nice feature!

dwave/cloud/solver.py
randomir: [Nope](https://docs.dwavequantum.com/en/latest/industrial_optimization/solver_nl_properties.html). I was surprised as well. Perhaps we should file a SAPI bug?

dwave/cloud/solver.py
randomir: It would be great to couple that change with the `problem -> job` change, so it's `run_job`, not `run_problem`...

dwave/cloud/solver.py
```diff
@@ -564,6 +583,28 @@ class BQMSolver(BaseUnstructuredSolver):
...
+            raise RuntimeError(
+                "dimod package with support for DiscreteQuadraticModel required."
+                "Re-install the library with 'dqm' support.")
+
+        bqm = BinaryQuadraticModel.from_ising({0: 1}, {})
+        params = {}
+        try:
+            params.update(time_limit=self.properties['minimum_time_limit'][0][1])
+        except (KeyError, IndexError):
+            pass
```
thisac, line 604: Do we expect `minimum_time_limit` to not be consistent in shape? Otherwise I'd think an `if self.properties.get('minimum_time_limit')`, etc. would make more sense.

dwave/cloud/solver.py
```diff
@@ -1232,6 +1343,64 @@ def sample_bqm(self, bqm, label=None, **params):
...
+        method, together with a set of required parameter values.
+        """
+        problem = ({min(self.nodes): 0}, {}, 0.0)
+        params = {}
+        return problem, params
+
+    def sample_problem(self,
+                       problem: tuple[list | dict[int, float],
+                                      dict[tuple[int, int], float],
+                                      float],
```
thisac, line 1358: Not too fond of this type-hint (aesthetically). Is it used elsewhere? Worth declaring outside of `sample_problem`?

dwave/cloud/solver.py
```diff
@@ -1232,6 +1343,64 @@ def sample_bqm(self, bqm, label=None, **params):
...
+            upload_params:
+                Optional upload/encode parameters, not used by structured solvers.
+
+            **sample_params:
+                Sampling parameters, solver-specific.
+
+        Returns:
+            Response in a future.
+
+        Note:
```
thisac, line 1393: Do we not prefer using Sphinx note directives `.. note:`?

thisac (approved):
LGTM overall. Thanks @randomir !

dwave/cloud/solver.py
randomir: With APIs, one should expect the unexpected! :rofl: 

Currently, solvers use `minimum_time_limit: list[list[number]]` and `minimum_time_limit_s: number`. That probably won't change, but in this case I chose EAFP over LBYL as it seemed simpler than checking for existence, then type, then size along the first axis, then second... And we need to check it, at least for now, because solver properties are not validated on client side after parsing.

dwave/cloud/solver.py
randomir: We _could_ pull this type from dimod, but dimod is not a required dependency, and we use it here only once. We can consider type aliasing on subsequent occurrence.

dwave/cloud/solver.py
randomir: We used `Note:` historically, so it's locally consistent. I agree the sphinx directive is semantically clearer, but I don't see other pros here. They're rendered the same (as admonition box) in Ocean docs, and I even find `Note:` a bit easier on the eye (and consistent with `Args:`, `Returns:`, etc.)

PR #755, opened 2026-07-30
Update `QCDLSolver` to accept a pydantic model

randomir:


setup.py
```diff
@@ -38,6 +38,9 @@
     # nlm support
     'nlm': ['dwave-optimization>=0.1.0,<0.5', 'numpy>=1.20.0'],
 
+    # qcdl (type checking) support
+    'qcdl': ['dwave-gate>=0.4'],
```
thisac, line 42: This will fail until we've made the release, no?

dwave/cloud/solver.py
```diff
@@ -1772,6 +1776,12 @@ def estimate_qpu_access_time(self,
         return sampling_time + programming_time
 
 
+""":class:`dwave.gate.qcdl.qcdl_models.Qcdl`, or a compatible type."""
+QCDLLike: TypeAlias = "Qcdl" \
```
thisac: Tangential, but do we still want to use `TypeAlias` even though it has been [deprecated](https://docs.python.org/3/library/typing.html#typing.TypeAlias)? I don't think it's going to be removed anytime soon, and there seem to be some debates around its usefulness compared to the recommended `type` statement (https://docs.python.org/3/reference/simple_stmts.html#type)

dwave/cloud/solver.py
```diff
@@ -1772,6 +1776,12 @@ def estimate_qpu_access_time(self,
...
+                      "| abc.Mapping[str, Any]" \
+                      "| abc.Callable[[], Qcdl | abc.Mapping[str, Any]]"
```
thisac: Should a ever expect a callable that outputs a mapping?

thisac (approved):
LGTM

setup.py
randomir: No. That's just an optional extras, not used anywhere, afaik.

dwave/cloud/solver.py
randomir: Yes. Until we drop Python<3.12, because `type` was introduced only in 3.12. So, `TypeAlias` remains the only way if you want to support older Pythons.

dwave/cloud/solver.py
randomir: That's basically the "old" flow, where `@qcdl`-decorated function was returning a dict. I see no harm in still supporting this case. If you look at the encoder below, you'll see it's cascaded, so supporting this case is free.

dwave/cloud/solver.py
```diff
@@ -1812,37 +1832,43 @@ def minimal_problem(self) -> tuple[dict, dict]:
...
 
         Args:
-            qcdl (dict/str):
+            qcdl:
                 A quantum circuit in a Quantum Circuit Description Language
-                (QCDL) dict, or a reference to one (Problem ID returned by the
-                :meth:`.upload_qcdl` method).
+                (QCDL) dict, model (:class:`~dwave.gate.qcdl.qcdl_model.Qcdl`),
+                a callable that returns one, or a reference to an uploaded
+                problem (Problem ID, as returned by the :meth:`.upload_qcdl`
```
qci-amos: it might be nice to define a type alias for Problem ID

dwave/cloud/solver.py
```diff
@@ -1789,7 +1799,17 @@ class QCDLSolver(BaseUnstructuredSolver):
     _handled_problem_types = {"qcdl"}
     _handled_encoding_formats = {"binary-ref"}
 
-    def _encode_problem_for_upload(self, qcdl, **kwargs):
+    def _encode_problem_for_upload(self, qcdl: QCDLLike, **kwargs):
+        if callable(qcdl):
+            qcdl = qcdl()
```
qci-amos: A while ago, I allowed my compiler to accept a `callable`, but later I regretted it and have since ripped it out. I recommend not supporting this.

I think It's better to force the user to create the `Qcdl` **before** they use your API because it can help differentiate:
1. if it's their bug or ours
2. if it's their generator that's slow or our processing

qci-amos (approved):
honestly, I could be happy with it ONLY accepting the `Qcdl` type! I think it's reasonable to force a user to live by its validation rather than letting them try to hack around the validation.

dwave/cloud/solver.py
thisac: Good point. I missed considering that `type` was introduced at the same time.

dwave/cloud/solver.py
randomir: That's a good point!

However, the cloud-client is a low-level library, intended mostly for advanced use(r) cases. We have higher-level interfaces in Ocean intended for general use. We'll add two higher-level APIs for submitting QCDLs to Leap -- one in the Qiskit plugin, and one in dwave-gate, similar to our Ocean samplers (where solver selection, QCDL encoding *and* decoding will be abstracted away). Those can be rigid, or even raise warning in the callable case.

Here, on the low-level, I'd rather err on the side of flexibility.

qci-amos: Yea, I added support for this when I too believed this flexibility would be useful. It just felt natural based on how the decorator works: why force the user to add those two extra parenthesis? but metaprogramming is different and not familiar to many users. There are significant sources of errors in the qcdl python the user writes that wouldn't raise until from inside your API. This means that anybody who takes this approach is going to unnecessarily and confusingly see your code in their stacktrace for many things that are completely their bugs. It can get especially nasty if the qcdl is actually an object and implement `__call__` as the means for generating the qcdl because now things are stateful. My opinion was changed based on years of experiences working with physicists who have extraordinarily complex logic used when generating their qcdls.

Furthermore, it's not common, but there are cases where the qcdl might require args/kwargs, so this isn't even the fully flexible solution anyway.

qci-amos (approved):

dwave/cloud/solver.py
randomir: I wasn't aware the decorator might require additional args. What's the use case for that?

I'll remove the callable option.

randomir: Removed in d868a931e39c29763e466d2764202546f3fd481c. 

Optional args make a good argument. Catering to bad programmers (even if they are physicists) does not. :slightly_smiling_face:

qci-amos: My view is that supporting `callable` is doing the catering to poor programming...

anyway, the use case for args/kwargs is simply that some folks want the same python code to generate different qcdls. This can be handled by having a stateful instance have `__call__` (i.e., parameters can slip in via `self`) but the simplest is to allow the decorated function to accept parameters directly.

PR #758, opened 2026-09-01
Accept additive SAPI changes

randomir:
Future-proof the client pydantic models to accept additive changes of SAPI responses.

Also, add the missing `QCDL` problem type omitted in #745.

**AI Generation Disclosure**

Used Claude to brain-storm the open enum mixin.


dwave/cloud/api/constants.py
```diff
@@ -20,7 +20,26 @@
 DEFAULT_API_RESPONSE_VERSION = '1.0.0'
 
 
-class ProblemStatus(str, enum.Enum):
+class _OpenEnumMixin:
```
thisac: Since this mixin only works is `cls` is of type `str`, perhaps better to have the str subclass be moved up a level?

```suggestion
class _OpenStringEnumMixin(str):
```

Or make it into an enum-type class and not a mixin at all?


```suggestion
class _OpenStringEnum(str, enum.Enum):
```

dwave/cloud/api/constants.py
```diff
@@ -20,7 +20,26 @@
...
+
+    Instead of failing when we receive an unknown enum from SAPI, we'll create
+    a its enum value on the fly.
+
+    See: protobuf's `open enum <https://protobuf.dev/programming-guides/enum/>`_.
+    """
+
+    @classmethod
+    def _missing_(cls, value):
+        member = str.__new__(cls, value)
```
thisac, line 36: Or generalize this.
```suggestion
        member_type = cls._member_type_
        if member_type is object:
            member = object.__new__(cls)
        else:
            member = member_type.__new__(cls, value)
```

thisac (approved):
LGTM. Just one minor comment.

dwave/cloud/api/constants.py
randomir: Sure, I can clarify the mixin's name. I didn't care much here about the name since it's a one-off private, internal class used only from within the constants module.

I will keep it as a mixin, though, since that seems to be the suggested pattern: [`EnumName([mixin_type, ...] [data_type,] enum_type)`](https://github.com/python/cpython/blob/f6842ddae6b0c744b89c33b9bb088d7c4b3eb8d3/Lib/enum.py#L943-L944).

